PR #4, opened 2021-02-25
Add README.md

timflannagan:
Mainly just a test for whether tide has been correctly configured in the release repository.


timflannagan:
This was mainly just a check for whether tide was setup correctly.

/close

timflannagan:
It looks like we're missing a couple of different plugins that I'm used to have in different repositories.

- close
- verify-owners

PR #7, opened 2021-02-25
Add root .gitignore

timflannagan:


dinhxuanvu:
Is there a reason we don't copy the upstream .gitignore over?

timflannagan:
I don't think the actual syncing mechanisms are working at the moment. There are a couple of things that I need to square away first, like getting the vendoring working correctly, but I'm fine with grabbing the OLM .gitignore over and updating this to use that instead.

dinhxuanvu (approved):
/lgtm

PR #9, opened 2021-02-25
staging: Sync the vendor dependencies

timflannagan:
There's currently a mismatch between what's defined in the go.mod and what's present in the go.sum, vendor/modules.txt which is causing tests to fail when attempting to add minimal prow jobs.

This is a precursor to #8 to also allow us to expand the prow tests in parallel to fixing the vendor directory.


dinhxuanvu (approved):
/lgtm

PR #10, opened 2021-03-01
Update the CI registry in the registry Dockerfile and copy OLM_VERSION to root directory

timflannagan:
The operator-registry Dockerfile currently references images that are now migrated to the registry.ci.openshift.org registry instead of registry.svc.ci.openshift.org which is failing image tests when attempting to grab that particular Go manifest tag.

Copy the OLM_VERSION from the OLM staging repository to the root repository as the OLM Dockerifle is expecting that file is present. There are a couple of different ways we can handle this, including updating the path to this file, but this kind of changes is a workaround anyways to get a base set of prow tests working so I don't think how we get access to this file is all that important at the moment.


ankitathomas (approved):
/lgtm

PR #11, opened 2021-03-01
pull required tags only and add commit message for pull script

ankitathomas:
Fetch only the specified tag for upstream sync, add the squashed commit list to the commit message


timflannagan (commented):
Can you split this into multiple commits on what's being changed on the script side vs. what's being pulled in after updating the staging dependencies?

go.mod
```diff
@@ -14,9 +14,9 @@ require (
 	github.com/onsi/gomega v1.10.3
 	github.com/openshift/api v0.0.0-20200331152225-585af27e34fd
 	github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0
-	github.com/operator-framework/api v0.5.1
+	github.com/operator-framework/api v0.5.0
 	github.com/operator-framework/operator-lifecycle-manager v0.0.0-00010101000000-000000000000
-	github.com/operator-framework/operator-registry v1.13.6
+	github.com/operator-framework/operator-registry v0.0.0-00010101000000-000000000000
```
timflannagan, line 19: This looks off me to me. Any reason we're not using the correct semantic tag?

go.mod
```diff
@@ -14,9 +14,9 @@ require (
 	github.com/onsi/gomega v1.10.3
 	github.com/openshift/api v0.0.0-20200331152225-585af27e34fd
 	github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0
-	github.com/operator-framework/api v0.5.1
+	github.com/operator-framework/api v0.5.0
```
timflannagan, line 17: Same here - this looks like a relatively stale API version.

staging/operator-registry/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/operator-framework/operator-registry
 
-go 1.15
+go 1.13
```
timflannagan, line 3: What version of the registry are we grabbing here? I know master has been bumped to 1.15 and I removed the apiserver dependency in a recent PR.

scripts/add_repo.sh
```diff
@@ -50,10 +51,27 @@ if [ ! -d "$remote_dir" ]; then
 
 	# add the subtree
 	git remote update $remote_name
-	ref=$(git show-ref remotes/$remote_name/master -s)
-	git subtree add --prefix="$remote_dir" "$remote_name" --squash master
-	sh -c "cd $remote_dir && go mod edit -replace $downstream_repo=../../"
-	git add $remote_dir/go.mod
+	ref=$(git show-ref remotes/$remote_name/${target_branch} -s)
```
timflannagan, line 54: I wasn't able to find anywhere this `ref` variable is being used.

timflannagan: (Not a big deal right now just something that I noticed)

go.mod
ankitathomas: This is for the replaces section, so we'd know we're pointing to the staged operator-registry repo instead of a vendored version.

scripts/add_repo.sh
```diff
@@ -50,10 +51,27 @@ if [ ! -d "$remote_dir" ]; then
...
+								go mod edit -replace $staged_mod=../$staged_dep"
+		grep "$new_mod" $staging_dir/$staged_dep/go.mod && sh -c "cd $staging_dir/$staged_dep && \
+								go mod edit -require $new_mod@v0.0.0-00010101000000-000000000000 && \
+								go mod edit -replace $new_mod=../$remote_name && \
+								go mod vendor && \
+								git add go.mod go.sum vendor"
+	done
+
+	sh -c "cd $remote_dir && \
+		go mod edit -replace $downstream_repo=../.. && \
```
timflannagan, line 71: Minor nit: I believe we likely want to avoid hardcoding that `../..` directory path. We may want to create something like a `ROOT_DIR` that points to the root directory path by grabbing the current working (e.g. `scripts`) directory and appending a `../` to that.

scripts/add_repo.sh
```diff
@@ -76,10 +94,13 @@ if $add_subtree ; then
 	# push to subtree dir
 
 	FORK_REMOTE=${FORK_REMOTE:-origin}
+	git checkout ${current_branch}
+	git merge --squash -s recursive -X theirs -m "Sync upstream $upstream_remote_name $target_branch" ${temp_branch}
```
timflannagan, line 98: :+1:

staging/operator-registry/go.mod
ankitathomas: This is the latest tagged version as of now, v1.16.1. master tracking master could pull in unstable code to downstream master.

go.mod
timflannagan: Gotcha, yeah that makes more sense the more I think about it.

scripts/add_repo.sh
ankitathomas: That's a good point, it would help with directory structure changes

go.mod
ankitathomas: This should be the placeholder `v0.0.0-00010101000000-000000000000` version, just to indicate the submodule status of the dependency

staging/operator-registry/go.mod
timflannagan: I was under the impression we could simply pin against known release branches (e.g. release-4.8 in this case) instead of releases, but that method wouldn't scale to subsequent releases once the upstream/downstream split has happened and upstream has its own release cadence, so this method makes more sense to me.

go.mod
timflannagan: Any reason we wouldn't be able to use `v0.0.0` instead? I'm not particularly worried, either way, but I think that's a bit clearer to me that's using a local package reference vs. the pseudo-semantic tag that Go spits out when it can't find a proper semantic tag.

go.mod
timflannagan: I need to wrap my head around whether we care about differing API tags between the registry and OLM components. Is it sufficient to how those overall packages share a single API version, or is it possible down the line that we'll need something like v0.5.x for the registry and v0.6.x for OLM?

PR #12, opened 2021-03-03
Update helper scripts and root Makefile

timflannagan:


scripts/unit.sh
```diff
@@ -0,0 +1,14 @@
...
+set -o pipefail
+
+: "${WHAT:?}"
+
+ROOT_DIR=$(dirname "${BASH_SOURCE[0]}")/..
+TARGET_NAME=${TARGET_NAME:="unit"}
+
+pushd "${ROOT_DIR}/staging/${WHAT}"
+make ${TARGET_NAME}
+popd
```
timflannagan, line 14: This is a bit awkward but this should work when calling that `unit` driver make target in the CI prow configurations.

ankitathomas (approved):
/lgtm

PR #13, opened 2021-03-04
Migrate towards having a single module per repository

timflannagan:
Update the repository structure and remove any nested go.mod,go.sum,vendor references.

Remap the existing package imports in the staging directories to use the root module + "staging" as the prefix.


timflannagan:
We're likely going in a different direction, but holding for now before closing.

/hold

PR #18, opened 2021-03-16
Update repository layout

timflannagan:
Update the repository layout and remove all nested vendor directories.

Remove any replace directives in the staging go.mod files.

Introduce a cmds root package that imports the staging operator-framework repositories.

Remove the mirrored cmd,pkg staging packages.

Update the Makefile and add basic build and testing targets.


timflannagan:
Hmm, I realize this is missing updated root Dockerfile(s) that assume the build context is in the root directory, and not the staging equivalent.

/hold

Makefile
```diff
@@ -1,6 +1,23 @@
 SHELL := /bin/bash
 ROOT_DIR:= $(patsubst %/,%,$(dir $(realpath $(lastword $(MAKEFILE_LIST)))))
 
+GO_BUILD_OPTS := -mod=vendor
+GO_BUILD_TAGS := -tags "json1"
+
+GIT_COMMIT := $(or $(SOURCE_GIT_COMMIT),$(shell git rev-parse --short HEAD))
+OPM_VERSION := $(or $(SOURCE_GIT_TAG),$(shell git describe --always --tags HEAD))
```
ankitathomas, line 8: If we're building from release branches, we won't have tags to get the version from.

ankitathomas (approved):
/lgtm

timflannagan:
There's a couple of things I'd like to tackle in this PR, but I won't hold up other work that can be done in parallel.

/hold cancel

Makefile
timflannagan: Yeah, that's a good point, and I have that same thing tracked locally in my notes. I think I'd prefer if we instead shift towards attempting to have this information stored as a file in the repository somewhere as we currently do with that tracked file in the scripts (iirc?) directory. That way, we don't have to rely on portions of the root .git repository when building the container images, and instead read from those file(s) to get the registry/OLM version information.

PR #19, opened 2021-03-16
Update the base image for the operator-registry image

timflannagan:
Update the base image tag to `rhel-8-golang-1.15-openshift-4.8` to match
what we have specified in the OLM base image as well. This allows us to
only have to override a single image input in the downstream OLM prow
repository configuration.


timflannagan:
/retitle Dockerfile: Update the Golang operator-registry base image tag

PR #20, opened 2021-03-18
Add additional make targets and update Dockerfile(s)

timflannagan:
Update the root Makefile and add additional test/build/etc. targets.

Add a helper script that is responsible for running the registry and/or OLM staging test e2e packages.

Re-sync the vendor directory after merging a change that updated one of the staging packages, to verify that the root vendor folder got updated with that change, and reverting that staging change without re-updating the root vendor directory.

Update the Dockerfile(s) and copy the entire source tree. Note: this is something we should avoid doing in the future, but this will allow us to easily build binaries for the time-being and add additional prow checks so we can get some actual CI feedback on subsequent PRs to this repository.


ankitathomas (approved):
/lgtm

PR #22, opened 2021-03-19
Update root and nested modules

timflannagan:
Update the root and nested modules and use the local repository content.

Use fake pseudo-versions for the operator-framework-related repositories that are present in the monorepo.


timflannagan:
Hmm, I'm not sure why that ci/prow/unit check still exists. I removed that in the last release updates earlier today.

/test all

PR #23, opened 2021-03-19
Introduce the non-opm registry binaries to the cmds package

timflannagan:
Update the list of local command binaries and add the non-opm (e.g.
deprecated) registry command packages so we can build and test them
using the root vendor tree.

Update the Makefile and aliases for building the olm and registry binaries.
We'll likely want to cleanup these build and testing targets, but this should be
sufficient for now.


timflannagan:
These changes got piggybacked into another PR.

/close

PR #26, opened 2021-03-19
staging: Update OLM modules after recent changes

timflannagan:
Update the staging OLM modules so that running `make vendor` in that
directory can be successful.

Note: we may need to add a check that verifies that `make vendor && git
diff --exit-code` 
for _all_ of the nested modules, and not just the root repository.

Also, when I re-ran `make vendor` at the root, I keep getting extra
files added. We may need to encode some logic of removing any
staging/operator-registry/pkg/sqlite/migrations/*.db-journal files after
pulling down upstream changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,11 +72,14 @@ unit/api:
 unit: ## Run unit tests
 	$(ROOT_DIR)/scripts/unit.sh
 
+e2e:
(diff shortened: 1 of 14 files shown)


ankitathomas (approved):
/lgtm

PR #27, opened 2021-03-22
add crd and manifest generation

ankitathomas:
Generate crds and manifests from the staged api repo


scripts/generate_crds_manifests.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -eu -o pipefail
```
timflannagan, line 3:
```suggestion
set -eou pipefail
```

scripts/generate_crds_manifests.sh
```diff
@@ -0,0 +1,56 @@
...
+
+repo_root=$(git rev-parse --show-toplevel)
+cd ${repo_root}
+export GOFLAGS="-mod=vendor"
+YQ="go run ./vendor/github.com/mikefarah/yq/v3/"
+CONTROLLER_GEN="go run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen"
+ver=$(cat ./OLM_VERSION)
+docker pull quay.io/operator-framework/olm:v${ver}
+olmref=$(docker inspect --format='{{index .RepoDigests 0}}' quay.io/operator-framework/olm:v${ver})
```
timflannagan, line 12: This looks a bit fragile to me, but it's fine for a first iteration imo.

timflannagan:
Throwing a hold as that e2e-aws-olm isn't a required check, but we probably want to double-check the must-gather contains the updated manifests.

/lgtm
/hold

timflannagan (approved):
Looks like these changes propagated correctly. From the must-gather tar file, the CSV CRD contains the `spec.cleanup` property definition.

```
             cleanup:
                description: Cleanup specifies the cleanup behaviour when the CSV
                  gets deleted
                properties:
                  enabled:
                    type: boolean
                required:
                - enabled
                type: object
```

timflannagan:
Oh, this already has the lgtm label. I don't think we need to respin that this test again so I'm going to remove the hold.

/hold cancel

PR #33, opened 2021-03-31
Makefile: Use the full git commit hash when building binaries

timflannagan:
Update the root Makefile and copy-paste the GIT_COMMIT from the
staging/operator-lifecycle-manager/Makefile that checks if any of the
ART-defined environment variables are present in the build environment,
and falling back to the full git commit hash.

This is important as there's existing tests in openshift/origin that
verify that full commit hash is present when running the
version/--version command/flags for the OLM and registry binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,19 +2,26 @@ SHELL := /bin/bash
 ROOT_DIR:= $(patsubst %/,%,$(dir $(realpath $(lastword $(MAKEFILE_LIST)))))
 CONTAINER_ENGINE := docker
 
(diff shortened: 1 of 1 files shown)


timflannagan:
Link to a rehearsal that caught this when attempting to promote from the downstream monorepo instead of the upstream equivalents: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/17323/rehearse-17323-pull-ci-openshift-operator-framework-olm-release-4.9-e2e-gcp/1376992350034726912.

PR #35, opened 2021-04-05
Bug 1945584: Remove overrides for GOOS and GOARCH in cpb Makefile recipe.

benluddy:
This appears to have been a mistake during the downstream migration
and results in broken bundle unpack jobs on non-x86 hosts.


timflannagan:
Adding a hold so this PR doesn't automatically merge (need to update the existing tests so they're not optional anymore) and we get some CI feedback.

/lgtm
/hold

timflannagan:
Looks good to go.

/hold cancel

PR #37, opened 2021-04-06
Upstream sync for api

ankitathomas:
initial sync to copy all upstream commit history


OWNERS
```diff
@@ -1,18 +0,0 @@
...
-  - exdx
-  - awgreene
-  - hasbro17
-  - ankitathomas
-  - joelanford
-  - timflannagan
-reviewers:
-  - timflannagan
-  - ankitathomas
-component: "OLM"
```
timflannagan, line 18: It looks like this got picked up by accident. We'll need to revert this and remove the nested OWNERS file in the staging/api directory.

benluddy:
Can we filter out staging/*/{vendor,OWNERS} on a per-commit basis? If the runs ~once per upstream commit, there will be a significant number of commits that are just deleting these files, and if I checkout any given downstream commit, my working tree is probably going to include all of those files.

timflannagan:
Yeah good point, Ben. We already filter out nested vendor in the staging repositories in the root .gitignore, but we need to also include the OWNERS file there too.

OWNERS
timflannagan: Actually, let's revert the change to the root OWNERS file, but tackle updating the root .gitingore in a separate PR based on [Ben's comment](https://github.com/openshift/operator-framework-olm/pull/37#issuecomment-814330580).

ankitathomas:
We can switch to per-commit filtering for vendor/OWNERS, but the sync would be slightly longer. It wouldn't add too much of an overhead to the syncs if it's done nightly, but it will add quite a bit when branches sync after a while

benluddy:
> We already filter out nested vendor in the staging repositories in the root .gitignore

It doesn't look like that is working. Check out this commit, for example: https://github.com/openshift/operator-framework-olm/pull/37/commits/ceb12cf47669720fec0006caacbd7c5eb4e964cf.

OWNERS
ankitathomas: Good catch there, thanks Tim!

PR #39, opened 2021-04-06
staging/olm: Re-generate test fake package(s)

timflannagan:
Run `go generate ./...` at the root of the
staging/operator-lifecycle-manager repository.


timflannagan:
Need to verify that the unit-olm check is green before merging.

/hold

timflannagan:
Looks like we're good to go

/hold cancel

ankitathomas (approved):
/lgtm

PR #43, opened 2021-04-08
Makefile: Introduce verify target that checks for unstaged changes

timflannagan:
This already lives as a check in the openshfit/release repository as a
series of commands, but ideally this should live as a Makefile target
that we instead just call in this verify target in the `verify` prow job.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,6 +121,10 @@ sanity:
 manifests: vendor ## Generate manifests
 	./scripts/generate_crds_manifests.sh
 
+verify: vendor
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -121,6 +121,10 @@ sanity:
 manifests: vendor ## Generate manifests
 	./scripts/generate_crds_manifests.sh
 
+verify: vendor
```
njhale, line 124: looks pretty similar to the `sanity` rule a few lines up.

PR #48, opened 2021-04-09
manifests,scripts: Add single node developer release annotation

timflannagan:
This was missed when we were pulling down upstream OLM changes.

- CP the OLM upstream commit that introduced this.
- Fix the scripts/generate_crd_manifests.sh bash script that points to a missing OLM_VERSION file.
- Re-generate the YAML manifests.
- Introduce the upstream OLM add_release_annotations bash script.

### DIFF
--- a/scripts/add_release_annotation.sh
+++ b/scripts/add_release_annotation.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# TODO(tflannag): Add ability to override yq binary as either a parameter
+# or an environment varaibles like upstream OLM does.
+# For now, you can get the yq binary by running `go install ./vendor/github.com/mikefarah/yq/v3`
+# from the root repository and moving that binary into $PATH.
+chartdir=$1
+
+for f in $chartdir/*.yaml; do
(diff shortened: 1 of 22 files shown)


timflannagan:
This is superseded by #49 

/hold

scripts/add_release_annotation.sh
```diff
@@ -0,0 +1,16 @@
...
+set -o nounset
+set -o pipefail
+
+# TODO(tflannag): Add ability to override yq binary as either a parameter
+# or an environment varaibles like upstream OLM does.
+# For now, you can get the yq binary by running `go install ./vendor/github.com/mikefarah/yq/v3`
+# from the root repository and moving that binary into $PATH.
+chartdir=$1
+
+for f in $chartdir/*.yaml; do
```
njhale, line 13: I know it's not really the topic of this PR to discuss design, but I don't really see a reason for this to be a script instead of a branch in our Helm templates; i.e. when `.Values.imagestream=true`: https://github.com/openshift/operator-framework-olm/blob/2df1ac17508e41b3191b04bc3fb00a30cbad681f/staging/operator-lifecycle-manager/deploy/chart/templates/0000_50_olm_99-operatorstatus.yaml#L1

timflannagan: Agreed - I may take some time in the next couple of weeks to tackle updating the layout of the charts + generate script(s).

PR #51, opened 2021-04-13
Bug 1948771: Revert "set packageserver replicas to 1 for single node"

timflannagan:
Cherry-pick of the upstream OLM commit: d74157da300a9874c7dd0f0200875038245e3d0e


ecordell:
/override e2e-upgrade

This reverts to a previously well-tested state, the one failed run doesn't include the error that this fixes, and not merging has widespread effects across all e2e-upgrade jobs for the org.

PR #53, opened 2021-04-14
Bug 1949294: fix environment variable merging from subscription.config

joelanford:
Downstream port of https://github.com/operator-framework/operator-lifecycle-manager/pull/2098


dinhxuanvu (approved):
/lgtm

timflannagan (commented):
You need to also run `make vendor` at the root so the changes made to the staging repositories are added to the root vendor tree.

dinhxuanvu (approved):
/lgtm

timflannagan:
/retest

```
failed to acquire lease: resources not found
--
```

jmazzitelli:
Just FYI: I provided my test case as a comment in the BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1949294#c6

I mention this because my use-case was not fixed by the stated workaround in that BZ - so it is possible what I am seeing is different and may or may not be fixed by this PR.

PR #60, opened 2021-04-19
Bug 1942522: Fix resolution error if inner entry doesn't provide a required API.

benluddy:
For bundle-to-bundle dependencies (as opposed to
subscription-to-bundle "dependencies"), candidates were filtered first
to only those that satisfy the constraint in question, then sorted in
channel order. Since the relative order of entries within a channel is
currently determined by replaces-distance from the channel head,
sorting would fail if an entry in the middle of a channel's replaces
chain did not satisfy the constraint.

For example, assume an operator "bar" has three available versions
within a single channel: bar-1, bar-2, and bar-3. If bar-1 and bar-3
provide the API "Bar", but bar-2 does not, then resolution will fail
for any operator that requires the API "Bar", because bar-1 and bar-3
are not connected by update edges without bar-2.

Now, all (package, channel, catalog) combinations satisfying a
constraint are enumerated and sorted before filtering to only those
entries that satisfy the constraint.

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/registry/resolver/cache.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/registry/resolver/cache.go
@@ -514,6 +514,12 @@ func WithLabel(label string) OperatorPredicate {
 	})
 }
 
+func WithCatalog(key registry.CatalogKey) OperatorPredicate {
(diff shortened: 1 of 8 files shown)


dinhxuanvu (approved):
/lgtm

PR #62, opened 2021-04-20
Bug 1934071: fix(catalog): Reduce namespace resync in resolution failure

dinhxuanvu:
Every resolution error will trigger namespace resync up to 8 times.
This sometimes leads to excessive resync even though the problem
is persistent.

NotSatisfiable error from resolver is often terminal and require
further manual intervetion. This type of error usually won't get
resolved by retrying resolution. As a result, if catalog encounters
this error, it should move on and doesn't force immediate resync.
The resync will happen naturally later when the default resync is
triggered.


benluddy:
/test unit-olm

That's a flaky test, tracked upstream in https://github.com/operator-framework/operator-lifecycle-manager/issues/2091.

PR #66, opened 2021-04-26
WIP: Set grace period for grpc pod deletion to default 30s

anik120:


staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/grpc.go
```diff
@@ -192,21 +192,35 @@ func (c *GrpcRegistryReconciler) EnsureRegistryServer(catalogSource *v1alpha1.Ca
 	// recreate the pod if no existing pod is serving the latest image
 	overwritePod := overwrite || len(c.currentPodsWithCorrectImage(source)) == 0
 
-	//TODO: if any of these error out, we should write a status back (possibly set RegistryServiceStatus to nil so they get recreated)
 	sa, err := c.ensureSA(source)
 	if err != nil && !k8serror.IsAlreadyExists(err) {
+		// TODO(anik120): introduce this condition reason in the api repo
```
timflannagan, line 197: We'll need to run `make vendor` to propagate any changes made to the Go packages in the relevant staging/* repositories.

anik120: Right, introduced this here as a trailer for what the plan is and to get the plan reviewed/approved. Will `make vendor` the api repo once the idea is approved.

staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/grpc.go
```diff
@@ -408,7 +419,7 @@ func imageID(pod *corev1.Pod) string {
 
 func (c *GrpcRegistryReconciler) removePods(pods []*corev1.Pod, namespace string) error {
 	for _, p := range pods {
-		err := c.OpClient.KubernetesInterface().CoreV1().Pods(namespace).Delete(context.TODO(), p.GetName(), *metav1.NewDeleteOptions(1))
+		err := c.OpClient.KubernetesInterface().CoreV1().Pods(namespace).Delete(context.TODO(), p.GetName(), metav1.DeleteOptions{})
```
timflannagan, line 422: I think before we merge this, I'd like to see whether we want a package-level constant for the default value (e.g. `metav1.DeleteOptions{}`) that has a comment that links to both of those BZs. Does that sound reasonable?

staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/grpc.go
timflannagan: Fair enough, but we build our binaries/images from the root vendor tree so it's possible that these changes won't get propagated to the CI feedback.

staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/grpc.go
timflannagan: Oh yeah, that's a good point - I'm not 100% sure whether we'd want a comment to that in upstream OLM (I see the argument for both sides), but that's a discussion we can have once we determine whether these fixes close out the attached BZ anyways.

anik120:
run 1: e2e-gcp is failing on other current usual suspect: [sig-instrumentation][Late] Alerts shouldn't report any alerts in firing or pending state apart from Watchdog and AlertmanagerReceiversNotConfigured and have no gaps in Watchdog firing
/retest

staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/grpc.go
timflannagan: We don't have a concrete downstream-specific strategy yet, namely because we don't have a mechanism that verifies that patches introduced downstream will be retained when pulling in upstream changes. In general, though, the same workflow would apply to any PR that changes the staging package code, is that we need to run `make vendor` so those changes propagate to the root vendor/ (which is why we check for unstaged changes in the verification check).

timflannagan: @anik120 The order of operations doesn't matter too much here as regardless of pulling in changes made upstream, or introducing downstream-only changes, we pull Go package changes to their corresponding staging file path, and re-generate the root vendor tree. We have setup the [root modules to use the local filesystem path](https://github.com/openshift/operator-framework-olm/blob/master/go.mod#L58-L61):

```
	// use staged repositories
	github.com/operator-framework/api => ./staging/api
	github.com/operator-framework/operator-lifecycle-manager => ./staging/operator-lifecycle-manager
	github.com/operator-framework/operator-registry => ./staging/operator-registry
```

And introduced a [`cmds` package at the root](https://github.com/openshift/operator-framework-olm/blob/master/cmds.go) so the package imports are correctly remapped, without having to do that remapping ourselves:

```go
package cmds

import (
	_ "github.com/operator-framework/operator-lifecycle-manager/cmd/catalog"
	_ "github.com/operator-framework/operator-lifecycle-manager/cmd/olm"
	_ "github.com/operator-framework/operator-lifecycle-manager/cmd/package-server"

	_ "github.com/operator-framework/operator-registry/cmd/configmap-server"
	_ "github.com/operator-framework/operator-registry/cmd/initializer"
	_ "github.com/operator-framework/operator-registry/cmd/opm"
	_ "github.com/operator-framework/operator-registry/cmd/registry-server"
)
```

timflannagan: And re-reading my last comment, we're technically double vendoring the OLM/API/registry packages, both in the root vendor (which is what we built from), and the staging/* repositories. That pattern is a bit strange, and it was mainly to get an MVP working sooner rather than later with the looming 4.8 feature freeze deadline, so we updated that ci/prow/verify check to ensure that when we run `make vendor` we have no unstaged changes, so we don't run into situations where the root vendor and staging/* packages are out of sync.

anik120: Gotcha, pushed changes from `make vendor` in with the commit.

PR #69, opened 2021-05-03
WIP: Exclude controlplane components from the guest cluster

ecordell:
This adds an annotation to resources that shouldn't be deployed in guests (because they will live in the control plane).

This is for testing, I will follow up with changes in the upstream `deploy` and pull that into this PR before merging.

### DIFF
--- a/manifests/0000_50_olm_00-namespace.yaml
+++ b/manifests/0000_50_olm_00-namespace.yaml
@@ -8,6 +8,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
   labels:
     openshift.io/scc: "anyuid"
     openshift.io/cluster-monitoring: "true"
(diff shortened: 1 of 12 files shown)


manifests/0000_50_olm_00-namespace.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
```
bparees, line 11: @ecordell i'm going to let the CVO keep creating this one, so we don't need to exclude it.

manifests/0000_50_olm_13-operatorgroup-default.yaml
```diff
@@ -17,6 +17,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
```
bparees, line 20: i'm not sure why this one was being excluded either, i don't think it should be.

ecordell:
This is no longer needed, though something similar might be needed in the future depending on how hypershift manifests will be defined

PR #71, opened 2021-05-05
Bug 1957386: Move config serve and validate to opm alpha

dinhxuanvu:


scripts/tracked
```diff
@@ -1,3 +1,4 @@
 operator-lifecycle-manager git@github.com:operator-framework/operator-lifecycle-manager.git master c480add342c70a38b98f2be7f8259261e4af01ba
 operator-registry git@github.com:operator-framework/operator-registry.git master 88eda1a2c6dab32c5a049a286259f3c0f713587e
 api git@github.com:operator-framework/api.git master 01e8bf8bd934a18aef1d22e9e4c523f293b2a633
+operator-registry git@github.com:operator-framework/operator-registry.git master e29651f8ecf77b32fb51491bbc395ebb2e527161
```
timflannagan, line 4: This is a duplicate of an existing entry. We're not currently maintaining this file it looks like.

staging/operator-lifecycle-manager/go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/onsi/gomega v1.10.2
 	github.com/openshift/api v0.0.0-20200331152225-585af27e34fd
 	github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0
-	github.com/operator-framework/api v0.0.0-00010101000000-000000000000
+	github.com/operator-framework/api v0.7.1
```
timflannagan, line 29: Hmm, this should still be using a fake pseudo tag as we replace this reference with the local filesystem package(s).

PR #84, opened 2021-06-02
Bug 1881522: packageserver CSV: add missing properties

vrutkovs:
Some properties are unset and being set to defaults. CVO is however unware of it, so it attempts to remove the fields which are unset in the manifest files.

This PR should prevent CVO from hotlooping on this manifest. Verified that its no longer happening:
```
$ curl -sL https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1400074947912536064/artifacts/launch/pods/openshift-cluster-version_cluster-version-operator-7d4cc845f6-pwfcm_cluster-version-operator.log | grep -oP "Updating \K.+ due to diff" | cut -d' ' -f1 | sort | uniq -c
    448 Deployment
     14 PriorityClass
```

### DIFF
--- a/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
+++ b/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
@@ -12,6 +12,9 @@ metadata:
     olm.version: 0.17.0
     olm.clusteroperator.name: operator-lifecycle-manager-packageserver
 spec:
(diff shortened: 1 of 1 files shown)


vrutkovs:
/test e2e-gcp

Install never started

PR #85, opened 2021-06-07
Bug 1958296: Surface API server warnings when applying plan steps.

benluddy:
Responses from the API server may include RFC2616 14.46 warnings (for
example, due to deprecation). Since the catalog-operator creates and
updates cluster resources on behalf of users, these warnings should be
surfaced to users.

When at least one warning is returned during the creation or update of
a step resource, catalog-operator now creates an Event with reason
"AppliedWithWarnings", involvedObject and fieldPath set to the plan
and step that generated the warning, respectively, and its message
containing the warning text.

In addition, a new metric, "installplan_warnings_total", counts the
number of steps that have been applied with warnings.


exdx:
/hold
generating downstream manifests

benluddy:
/hold

Test flakiness should go down once https://github.com/openshift/operator-framework-olm/pull/88 merges (affecting e2e-aws-olm), but e2e-gcp is failing due to a test of the alert trigger condition (https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_operator-framework-olm/85/pull-ci-openshift-operator-framework-olm-master-e2e-gcp/1401961122583547904).

anik120:
/hold

Looks like we're waiting on https://github.com/openshift/origin/pull/26211 too

PR #86, opened 2021-06-07
Bug 1965334: drop xattrs during unpack

njhale:
On some systems, creating files with certain extended attributes is a
restricted operation -- e.g. "security.compatibility" on SELinux -- and
causes bundle and index unpacking to fail for unprivileged users. To fix
this, we drop all xattrs from unpacked files before writing them.

Upstream-commit: 3bd849f97ca3b71cf0cfeea2fac8101302a3eeca

### DIFF
--- a/staging/operator-registry/pkg/image/containerdregistry/registry.go
+++ b/staging/operator-registry/pkg/image/containerdregistry/registry.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"regexp"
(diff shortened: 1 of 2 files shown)


timflannagan:
@njhale It looks like the commit references an upstream OLM commit but these changes are only touching the registry codebase.

timflannagan:
/retitle Bug 1965334: drop xattrs during unpack

PR #87, opened 2021-06-07
WIP: Introduce a PackageServer Deployment controller

timflannagan:
Placeholder to have clusterbot create a release payload image that can be used to test upgrades.


staging/operator-lifecycle-manager/pkg/controller/operators/packageserver_controller.go
```diff
@@ -0,0 +1,165 @@
...
+	Namespace string
+	// TODO(tflannag): Should this be a custom struct instead?
+	// TODO(tflannag): Add a mutex here?
+	// TODO(tflannag): Add an event recorder?
+	HighlyAvailableMode bool
+}
+
+//+kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures/finalizers,verbs=update
```
njhale, line 64: nit: we can probably dial these back a notch -- and add some of the core types this needs to play with too.

staging/operator-lifecycle-manager/pkg/controller/operators/packageserver_controller.go
```diff
@@ -0,0 +1,165 @@
...
+		return []reconcile.Request{}
+	}
+
+	// TODO(tflannag): Build up a simple setter if we need a more concrete structure that
+	// holds this information, vs. a single field?
+	r.HighlyAvailableMode = true
+	topologyMode := infra.Status.ControlPlaneTopology
+	if topologyMode == configv1.SingleReplicaTopologyMode {
+		r.HighlyAvailableMode = false
+	}
```
njhale, line 143: We need a mutex here. This isn't thread safe.

staging/operator-lifecycle-manager/pkg/controller/operators/packageserver_controller.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, r.Client.Create(ctx, deployment)
+	}
+
+	// TODO(tflannag): Need to handle updating any non-replica high availability fields
+	// e.g. hard anti affinity, maxUnavailable, etc.
+	expectedReplicas := int32(defaultReplicaCount)
+	if !r.HighlyAvailableMode {
```
njhale, line 88: we need a mutex here.

staging/operator-lifecycle-manager/pkg/controller/operators/packageserver_controller.go
```diff
@@ -0,0 +1,165 @@
...
+		r.Log.Info("packageserver will be deployed in non-HA mode")
+		deployment.Spec.Replicas = pointer.Int32Ptr(1)
+	}
+
+	return deployment, nil
+}
+
+func (r *PackageServerReconciler) infrastructureHandler(obj client.Object) []reconcile.Request {
+	// TODO(tflannag): Is this thread safe? Do I need to instantiate a mutex in the reconciler struct?
+	// TODO(tflannag): Is this an abuse of the mapping handler function? Should we be altering state here?
```
njhale, line 125: If we miss an event on the Infrastructure resource, the controller can be easily left in an inconsistent state, which can lead to incorrect scaling when the packageserver's replicas are set.

I suggest making a client/cache call in the `Reconcile` function instead.

njhale (commented):
Thank you for putting this together.

Right now I'm mostly wondering if this would be better as a more general "SNO sidecar container". There's almost nothing OLM-specific, so conceivably, we could keep our changes out of staging entirely.

manifests/0000_50_olm_06-packageserver.apiservice.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  name: v1.packages.operators.coreos.com
+  annotations:
+    service.alpha.openshift.io/inject-cabundle: "true"
```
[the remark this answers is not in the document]
timflannagan: Correct, the PackageServer component(s) will be deployed at a run-level right below the OLM Operator component(s) in the current implementation. In previous iterations, the controller had startup logic (e.g. by instantiating a `Runnable` method) that would ensure the PackageServer Deployment resource would be created if it's not already present, before starting the actual controller as we're only watching that specific deployment through a predicate function that checks if the current obj.Name is equal to the PackageServer's Deployment metadata.name. If that resource doesn't exist, then we wouldn't be handling any events, and therefore wouldn't be able to react to HA expectations that are being exposed through an in-cluster API.

Now, the CVO is responsible for creating the underlying k8s resources that PackageServer requires, so we can make the assumption that when the controller has been started and we're reconciling events, that deployment already exists and we can focus on reconciling any changes based on that in-cluster API.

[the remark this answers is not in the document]
timflannagan: Nope, just an oversight as these manifests was built up for POC purposes and haven't been touched much since then. This annotation has since graduated to beta: https://docs.openshift.com/container-platform/4.7/security/certificates/service-serving-certificate.html#add-service-certificate-configmap_service-serving-certificate.

PR #91, opened 2021-06-09
Bug 1927340: Add OperatorCondition status sync and update operator upgradeable check

dinhxuanvu:
OperatorCondition controller will update its status to refect the
changes on its spec regarding the operator conditions that are reported
by the operators themselves. In turn, operators can read the status
to confirm if OLM has processed the spec changes.

OLM will not take actions on upgradeable condition if the status is
stale using ObservedGeneration and Generation check.

Upstream-commit: 4ae8df668dd098c204f757b8572fa608a9899bdc

### DIFF
--- a/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
+++ b/manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
@@ -74,6 +74,9 @@ spec:
           spec:
             strategy:
               type: RollingUpdate
+              rollingUpdate:
+                maxUnavailable: 1
+                maxSurge: 1
             replicas: 2
             selector:
               matchLabels:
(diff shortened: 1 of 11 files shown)


go.mod
```diff
@@ -10,11 +10,14 @@ require (
...
-	github.com/operator-framework/operator-lifecycle-manager v0.0.0-00010101000000-000000000000
-	github.com/operator-framework/operator-registry v0.0.0-00010101000000-000000000000
+	github.com/mitchellh/hashstructure v1.0.0
+	github.com/mitchellh/mapstructure v1.1.2
+	github.com/onsi/ginkgo v1.16.1
+	github.com/onsi/gomega v1.11.0
+	github.com/openshift/api v0.0.0-20200331152225-585af27e34fd
+	github.com/openshift/client-go v0.0.0-20200326155132-2a6cd50aedd0
+	github.com/operator-framework/api v0.9.2
+	github.com/operator-framework/operator-registry v1.13.6
```
timflannagan, line 20: This looks off - cherrypick individual commits downstream to the staging directory shouldn't produce a diff here as we're always remapping these imports to the local filepath.

timflannagan: Ah, I see what happened. Some of the changes here were introduced to the top-level pkg directory and not the staging path.

PR #94, opened 2021-06-14
manifests,staging: Add default values to the PackageServer CSV and regenerate manifests

timflannagan:
Update the staging/operator-lifecycle-manager helm charts for the
PackageServer CSV resource and add the defaults for various fields to
avoid the CVO hot looping when attempting to create this resource.

These changes were introduced directly into the root list of manifests
in https://github.com/openshift/operator-framework-olm/pull/84 and this
PR attempts to ensure those changes were also propagated to the helm
charts and regenerated to keep the manifest directory in sync.


timflannagan:
Going in a different direction to avoid making changes to the staging tree.

PR #95, opened 2021-06-15
Makefile,scripts: Introduce additional verification checks

timflannagan:
Introduce the scripts/check-staging-vendor.sh bash script that's
responsible for ensuring the API, operator-registry, and
operator-lifecycle-manager staging repositories have up-to-date go.sum
and go.mod files. This is namely helpful for validating that
cherrypicked commits have updated the nesting module dependencies when
pulling changes downstream.

Update the verify target in the root Makefile and add additional targets
responsible for ensuring that the CVO manifests (the root manifest
directory) have been properly rendered after changes made to the helm
chart(s), verifying the nested vendor contains an updated go.sum and
go.mod when pulling down changes.

### DIFF
--- a/scripts/generate_crds_manifests.sh
+++ b/scripts/generate_crds_manifests.sh
@@ -1,22 +1,25 @@
-#!/bin/bash 
+#! /bin/bash
 
-set -euo pipefail
+set -o errexit
+set -o nounset
+set -o pipefail
 
-repo_root=$(git rev-parse --show-toplevel)
-cd ${repo_root}
+ROOT_DIR=$(dirname "${BASH_SOURCE[0]}")/..
 export  GOFLAGS="-mod=vendor"
+
 YQ="go run ./vendor/github.com/mikefarah/yq/v3/"
 CONTROLLER_GEN="go run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen"
-ver=$(cat ./OLM_VERSION)
+HELM="go run helm.sh/helm/v3/cmd/helm"
 
+ver=$(cat ./OLM_VERSION)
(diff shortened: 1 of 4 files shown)


scripts/generate_crds_manifests.sh
```diff
@@ -36,24 +39,39 @@ done
...
+}
+
+update_csv() {
+   local csv=$1
+
+   yq w --inplace "${csv}" --tag '!!bool' 'spec.cleanup.enabled' false
+   yq w --inplace "${csv}" 'spec.customresourcedefinitions' {}
+   yq w --inplace "${csv}" --style="" 'spec.install.spec.deployments[0].spec.template.spec.containers[0].ports[0].protocol' TCP
+   yq w --inplace "${csv}" --style="" 'spec.install.spec.deployments[0].spec.template.metadata.creationTimestamp' null
+   sed -i "s/'{}'/{}/g" "${csv}"
```
timflannagan, line 69: This is needed as the `yq w --inplace "${csv}" 'spec.customresourcedefinitions' {}` command above outputs a `'{}'` as the default style. I wasn't able to find a way to do this natively using yq v3, as there weren't any non-default styles that were removing the quotes as needed, or specifying `--tag '!!map'` was being recognized as a custom tag, vs. explicitly casting to a map key. This should be fine in the short-term as this resource will be removed from CVO control (and thus from this root list of manifests) in 4.9+.

timflannagan:
Verification check passed - no need to take e2e leases for a non-code PR.

/override ci/prow/e2e-aws-console-olm
/override ci/prow/e2e-aws-olm
/override ci/prow/e2e-gcp
/override ci/prow/e2e-upgrade

PR #96, opened 2021-06-16
Sync upstreams - 4.9 master open (squashed)

njhale:
## Description

Squashed approach for syncing upstreams -- uses `./scripts/pull_upstream.sh`.

## Side Effects

- no recognizable commit history
- squashed commits contain changes already made via cherry-picks
- downstream-only dependency changes


njhale:
Going with a cherry-pick approach instead with the aim of retaining some semblance of history.

PR #99, opened 2021-06-17
Migrate the PackageServer CSV management to a secondary controller

timflannagan:
Introduce a downstream-only controller that is responsible for managing
the PackageServer CSV resource. Previously, this resource was created
and managed by the CVO. In order to comply with SNO expectations in
4.9+, the PackageServer CSV management needs to move to a second-level
operator that can react to high availability expectations exposed by the
in-cluster Infrastructure cluster singleton API.

This controller creates a filtered watch on the PackageServer CSV
resource, and a watch on the Infrastructure cluster singleton API.

When reconciling events, the main reconciliation logic ensures that the
current high availability expectations are present in the CSV resource,
letting the OLM Operator propagate that desired state to the underlying
PackageServer-related resources.

This typically means ensuring PackageServer is deployed with
only a single replica in SNO-enabled cluster environments (denoted
through the SingleReplica topology mode), else is deployed in the
default configuration.

In order to ensure that the PackageServer container image is inline with
the rest of the OLM-related container images, we can continue to use the
find-and-replace functionality that the CVO provides when inspecting the
manifests/image-references file at release payload build time, and when
starting up. This allows us to specify an environment variable to the
psm executable that matches the OLM image specified in the image
references manifest, letting the CVO handle the substitution to the
correct image, and then propagating the value of that environment
variable to the CSV resource when reconciling events.

Future work likely includes migrating from deploying the PackageServer
using the OLM-native CSV API. This would involve the psm executable
rendering and deploying a Deployment, APIService, Service, etc.
resources, which details the need to have a solidified migration
strategy given the CSV places an owner reference on the underlying
resources, and simply removing that CSV will result in a downtime and a
SLA API violation as the underlying resources would be deleted.


pkg/package-server-manager/controller.go
```diff
@@ -0,0 +1,192 @@
...
+			_ = ensureCSV(log, csv, r.getHighlyAvailableMode())
+			if err := r.Client.Create(ctx, csv); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	}
+
+	if !csv.ObjectMeta.DeletionTimestamp.IsZero() {
+		log.Info("requeuing after encountering a delete event")
+		return ctrl.Result{Requeue: true}, nil
```
njhale, line 91: I know there's also a way to introduce a wait period before the next requeue occurs. Maybe we should use this to cut down on churn? I think it should reduce the chance of queue backoff in case a deletion is slow for some reason.

pkg/package-server-manager/controller.go
```diff
@@ -0,0 +1,192 @@
...
+		if apierrors.IsNotFound(err) {
+			log.Info("csv does not exist attempting to create it")
+			csv, err = manifests.NewPackageServerCSV(r.Namespace)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			_ = ensureCSV(log, csv, r.getHighlyAvailableMode())
+			if err := r.Client.Create(ctx, csv); err != nil {
+				return ctrl.Result{}, err
+			}
```
njhale, line 85: I think we want to return immediately after a successful create, since we know that if the resource changes it will be enqueued again anyway.

pkg/package-server-manager/controller.go
```diff
@@ -0,0 +1,192 @@
...
+		Watches(&source.Kind{Type: &configv1.Infrastructure{}}, handler.EnqueueRequestsFromMapFunc(r.infrastructureHandler)).
+		Complete(r)
+}
+
+func ensureCSV(logger logr.Logger, csv *olmv1alpha1.ClusterServiceVersion, highlyAvailableMode bool) bool {
+	var modified bool
+
+	deploymentSpecs := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+	if len(deploymentSpecs) != 1 {
+		logger.Info("not processing csv with more than one nested deployment")
```
njhale, line 167: nit: more than one or zero?

pkg/package-server-manager/controller.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	if !csv.ObjectMeta.DeletionTimestamp.IsZero() {
+		log.Info("requeuing after encountering a delete event")
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	orig := client.MergeFrom(csv.DeepCopy())
+	if !ensureCSV(log, csv, r.getHighlyAvailableMode()) {
+		log.Info("no updates to the csv are necessary")
```
njhale, line 96: nit: will these logs show up in non-debug configurations? If so, I think this might end up generating a lot of noise.

pkg/package-server-manager/controller.go
```diff
@@ -0,0 +1,192 @@
...
+		Complete(r)
+}
+
+func ensureCSV(logger logr.Logger, csv *olmv1alpha1.ClusterServiceVersion, highlyAvailableMode bool) bool {
+	var modified bool
+
+	deploymentSpecs := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+	if len(deploymentSpecs) != 1 {
+		logger.Info("not processing csv with more than one nested deployment")
+		return false
```
njhale, line 168: IIUC, this means that a user can remove (or add) a deployment spec to the CSV and the packageserver manager will do nothing to correct the CSV. This is a change in behavior from what happens with CVO today (since it will stomp those changes).

My question is, do we want to diverge from CVO's behavior?

pkg/package-server-manager/controller.go
```diff
@@ -0,0 +1,192 @@
...
+func (r *PackageServerCSVReconciler) infrastructureHandler(obj client.Object) []reconcile.Request {
+	log := r.Log.WithValues("infrastructure", obj.GetName())
+
+	if obj.GetName() != infrastructureName {
+		return nil
+	}
+
+	var infra configv1.Infrastructure
+	if err := r.Client.Get(context.Background(), types.NamespacedName{Name: obj.GetName()}, &infra); err != nil {
+		return []reconcile.Request{}
```
njhale, line 117: why return `[]reconciler.Request{}` over `nil` here?

pkg/package-server-manager/controller.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (r *PackageServerCSVReconciler) infrastructureHandler(obj client.Object) []reconcile.Request {
+	log := r.Log.WithValues("infrastructure", obj.GetName())
+
+	if obj.GetName() != infrastructureName {
+		return nil
+	}
+
+	var infra configv1.Infrastructure
+	if err := r.Client.Get(context.Background(), types.NamespacedName{Name: obj.GetName()}, &infra); err != nil {
```
njhale, line 116: I'm a fan of serializing this `Get` in the `Reconcile` function instead of in the infrastructure handler. I believe doing so would remove any question of how the threadiness of the manager affects consistency.

pkg/package-server-manager/controller.go
timflannagan: Unfortunately, there's no notion of a debug handler attached to the log.Logger that was generated at scaffolding time by kubebuilder. I believe you can instead just specify the verbosity level, so maybe `log.V(5).Info(...)` would be more appropriate?

pkg/package-server-manager/controller.go
timflannagan: Is there a back-off time duration you had in mind? I'm wondering if 10s is sufficient, or whether that value should be extended. Maybe it's contingent on HA vs. non-HA if there are lesser API obligations with the latter cluster environments, and we can afford some minimal downtime.

pkg/package-server-manager/controller.go
timflannagan: That was previously a stub that was used to avoid blindly indexing into that array of deploymentSpec(s). I'm wondering whether we'd prefer to check if we're handling a CSV that contains more nested deploymentSpec than what we're expecting, we should just serialize the YAML manifest directly, and toggle the modified bool to true. That way, even if there are no modifications needed on the replica/rolloutStrategy configuration side-of-things, the CSV parameter will be assigned the new deploymentSpec. I think we'd like to avoid exposing a latch for users/controllers/etc. to make changes to this CSV that we don't support.

pkg/package-server-manager/controller.go
timflannagan: I guess there's also no real reason we have to use this logging library, instead of logrus, which seems to be the favorite when poking around the various OLM-related repositories.

cmd/package-server-manager/main.go
```diff
@@ -0,0 +1,101 @@
...
+	opts := zap.Options{
+		Development: true,
+	}
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+
+	setupLog := ctrl.Log.WithName("setup")
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), manager.Options{
+		Scheme:                  setupScheme(),
+		Namespace:               namespace,
+		MetricsBindAddress:      "0",
```
timflannagan, line 72: This is likely missing a TODO comment about re-enabling metrics unless there's any reason we wouldn't want to expose that functionality in the container.

timflannagan: Ah, interesting it looks like we also disable this currently for OLM as well looking at the staging/operator-lifecycle-manager/cmd/olm main package(s).

pkg/package-server-manager/controller.go
timflannagan: Updated to use another implementation.

pkg/package-server-manager/controller.go
timflannagan: Updated to a helper function that validates whether the deployment and individual containers match the expectations we have and if not, the parent site overrides the current CSV with the default YAML manifest configuration that gets passed to the ensureCSV function.

manifests/0000_50_olm_07-olm-operator.deployment.ibm-cloud-managed.yaml
```diff
@@ -24,6 +24,38 @@ spec:
...
+          - $(PACKAGESERVER_NAMESPACE)
+          - --metrics
+          - :8083
+          image: quay.io/operator-framework/olm@sha256:de396b540b82219812061d0d753440d5655250c621c753ed1dc67d6154741607
+          imagePullPolicy: IfNotPresent
+          env:
+          - name: PACKAGESERVER_NAME
+            value: packageserver
+          - name: PACKAGESERVER_IMAGE
+            value: quay.io/operator-framework/olm@sha256:de396b540b82219812061d0d753440d5655250c621c753ed1dc67d6154741607
```
timflannagan, line 44: A quick note on this as I didn't want to add a comment to the YAML manifest - the CVO currently exposes a way to override default container images through an image-references file, which is mounted in a component's container image, merged at release payload build time, and resolved at startup. In order to reference the packageserver container image without having to query on-demand for that information or saving that state to a k8s resource, we can use an environment variable that the CVO will essentially sed if the image specified matches the image detailed in the image-references file.

pkg/manifests/manifests.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	return &csv, nil
+}
+
+func WithName(name string) CSVOption {
+	return func(csv *olmv1alpha1.ClusterServiceVersion) {
+		csv.Name = name
+	}
+}
```
timflannagan, line 38: This may be an indication that we need to either continue in this direction for the fields we care about (namely HA-related fields, like `.spec.replicas`), or start gradually migrating towards using a go-template version of the CSV that's templated dynamically according to HA expectations exposed by in-cluster APIs.

cmd/package-server-manager/main.go
timflannagan: I ended up enabling this and moved the default port value to 8083 to avoid conflicting with the OLM Operator metric ports (8080-8082). Likely another reason why it would be easier to deploy this controller as a standalone deployment vs. a sidecar container, but we can revisit this approach later in the 4.9 development cycle.

pkg/manifests/csv.yaml
```diff
@@ -6,10 +7,6 @@ metadata:
   labels:
     olm.version: 0.18.3
     olm.clusteroperator.name: operator-lifecycle-manager-packageserver
-  annotations:
-    include.release.openshift.io/ibm-cloud-managed: "true"
-    include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
```
timflannagan, line 12: This diff looks off.

/hold

timflannagan: /hold cancel

njhale:
/approve

this looks good to me -- but we should get one more independent review before merging (as per our policy).

pkg/package-server-manager/controller.go
```diff
@@ -0,0 +1,148 @@
...
+			manifests.WithNamespace(csv.Namespace),
+			manifests.WithImage(image),
+		)
+		if err != nil {
+			return err
+		}
+		csv.Spec = tmp.Spec
+	}
+	if !ensureCSV(log, image, csv, highAvailabilityMode) {
+		log.V(3).Info("no further updates are necessary to the csv")
```
ecordell, line 115: nit: might want to indicate this log is for packageserver (or omit it altogether)

pkg/package-server-manager/controller.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *PackageServerCSVReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	predicateNameFilter := predicate.NewPredicateFuncs(func(object client.Object) bool {
+		return object.GetName() == r.Name && object.GetNamespace() == r.Namespace
+	})
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&olmv1alpha1.ClusterServiceVersion{}, builder.WithPredicates(predicateNameFilter)).
```
ecordell, line 145: predicates just filter the events the reconciler sees; the cache will still fill with all of the CSVs (including copied CSVs, etc)

instead you should set up a cache `Field` filter on the name/namespace of the CSV (similar to https://github.com/operator-framework/operator-lifecycle-manager/pull/2248/files but for the name, not a label)

that will ensure the server only sends you events for that object and the cache doesn't fill up with irrelevant CSVs

scripts/generate_crds_manifests.sh
```diff
@@ -89,8 +80,89 @@ spec:
       name: quay.io/operator-framework/configmap-operator-registry:latest
 EOF
 
+cat << EOF > manifests/0000_50_olm-06-psm-operator.deployment.yaml
```
ecordell, line 83: is there a way this can go into a real file? I'm missing why cat/heredoc is needed (if this was intentional)

ecordell (commented):
looking really good! just a couple of comments

scripts/generate_crds_manifests.sh
timflannagan: Nope, no real reason. I can tackle something nicer in a follow-up PR.

PR #102, opened 2021-06-23
Bug 1975432: Resolve InstallPlanStepAppliedWithWarnings alert after some time.

benluddy:
There is an error in the alert expression that causes the alert to
persist indefinitely. The intention of the alert was to fire for
several minutes to inform the administration of the warnings. The
appropriate response to the alert is to stop installing problematic
bundles going forward.

### DIFF
--- a/manifests/0000_90_olm_01-prometheus-rule.yaml
+++ b/manifests/0000_90_olm_01-prometheus-rule.yaml
@@ -33,7 +33,7 @@ spec:
     - name: olm.installplan.rules
       rules:
         - alert: InstallPlanStepAppliedWithWarnings
-          expr: sum(sum_over_time(installplan_warnings_total[5m])) > 0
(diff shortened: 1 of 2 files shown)


ecordell:
approved, I think shipping 4.8 without this will lead to a higher case load (the alerts may be alarming / confusing without this change, and they're likely to be triggered)

ecordell:
/override ci/prow/e2e-gcp

e2e-gcp is only failing due to a single known issue tracked here: https://bugzilla.redhat.com/show_bug.cgi?id=1975283 (a test that should be skipped is not)

ecordell:
/override ci/prow/e2e-gcp

e2e-gcp is only failing due to a single known issue tracked here: https://bugzilla.redhat.com/show_bug.cgi?id=1975283 (a test that should be skipped is not)

PR #104, opened 2021-06-25
Bug 1975432: Resolve InstallPlanStepAppliedWithWarnings alert after some time.

kevinrizza:
There is an error in the alert expression that causes the alert to
persist indefinitely. The intention of the alert was to fire for
several minutes to inform the administration of the warnings. The
appropriate response to the alert is to stop installing problematic
bundles going forward.

Upstream-commit: b61e9f68738f8fc4e48653e6eaa5663816a38019


benluddy:
/close

Duplicate of https://github.com/openshift/operator-framework-olm/pull/102.

PR #107, opened 2021-06-29
scripts: Avoid relying on the staging/operator-lifecycle-manager helm chart templates

timflannagan:
Refactor the scripts/generate_crds_manifest.sh bash script responsible
for generating a series of static YAML manifests that the CVO can
consume and apply when rolling out a release payload.

As we further decouple upstream OLM from housing OCP-specific
resources/logic/etc., we should avoid relying on the
staging/operator-lifecycle-manager helm chart templates as the primary
avenue to producing this static YAML manifests. Instead, we can directly
move any YAML manifest in the root manifest directory as we have little
need to inject dynamic data so no templating tool should be required
downstream.

TODO: avoid using the staging/operator-lifecycle-manager charts when
generating the CRDs manifests.

### DIFF
--- a/scripts/generate_crds_manifests.sh
+++ b/scripts/generate_crds_manifests.sh
@@ -5,55 +5,62 @@ set -o nounset
 set -o pipefail
 
 ROOT_DIR=$(dirname "${BASH_SOURCE[0]}")/..
-export  GOFLAGS="-mod=vendor"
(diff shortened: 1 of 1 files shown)


timflannagan:
Closing as this was updated as a part of the downstream 4.9 port PR.

PR #112, opened 2021-07-01
Implement Continuous Profiling

awgreene:


manifests/0000_50_olm_02-services.yaml
```diff
@@ -17,6 +17,10 @@ spec:
       port: 8081
       protocol: TCP
       targetPort: metrics
+    - name: https-pprof
+      port: 8443
+      protocol: TCP
+      targetPort: 8443
```
timflannagan, line 23: Unfortunately, we'll likely need to land #107 before we can directly modify the root list of manifests. You would need to update the staging/operator-lifecycle-manager helm chart templates to have these changes propagate to this directory. In the script responsible for generating the list of manifests, we always nuke the root manifests directly and replace it with a tmp directory that contained the rendered helm chart templates. The verification check will fail as `make manifests && git diff` will produce a diff as this PR introduced manifests, and updates existing manifests, which will removed due to the process outlined above.

timflannagan: And ideally, we wouldn't be making direct modifications to the staging/* directory paths as we're likely heading down the path of adding validation that ensures changes made to the staging directories are only from cherrypick-ed upstream commits.

awgreene: Changes to manifests are now integrated into the `make manifests` command. No files in the `staging/*` directory were hurt as a part of this change.

manifests/0000_50_olm_07-collect-profiles.cronjob.yaml
```diff
@@ -0,0 +1,43 @@
...
+              command:
+                - bin/collect-profiles
+              args:
+                - -n
+                - openshift-operator-lifecycle-manager
+                - --config-path
+                - /etc/config
+                - --client-ca
+                - /var/run/secrets/serving-cert
+                - olm-operator-heap-:https://olm-operator-metrics:8443/debug/pprof/heap
```
benluddy, line 30:
```suggestion
                - olm-operator-heap-:https://olm-operator-metrics:8443/debug/pprof/heap
                - catalog-operator-heap-:https://catalog-operator-metrics:8443/debug/pprof/heap
```

operator-lifecycle-manager.Dockerfile
```diff
@@ -14,6 +14,8 @@ RUN mkdir -p .git/objects
 COPY . .
 RUN make build/olm bin/cpb
 
+RUN chmod +x /build/bin/collect-profiles
```
benluddy, line 17: I'm surprised that this would be necessary. What's special about how this binary is being produced?

manifests/0000_50_olm_07-collect-profiles.cronjob.yaml
```diff
@@ -0,0 +1,43 @@
...
+  namespace: openshift-operator-lifecycle-manager
+spec:
+  schedule: "*/15 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: collect-profiles
+          containers:
+            - name: collect-profiles
```
benluddy, line 18: This pod template needs one of the allowed priority classes:
```
4 pods found with invalid priority class (should be openshift-user-critical or begin with system-): openshift-operator-lifecycle-manager/collect-profiles-27107640-b6f84 (currently "") openshift-operator-lifecycle-manager/collect-profiles-27107640-ls2rz (currently "") openshift-operator-lifecycle-manager/collect-profiles-27107655-sg82n (currently "") openshift-operator-lifecycle-manager/collect-profiles-27107655-tdmzg (currently "")
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_operator-framework-olm/112/pull-ci-openshift-operator-framework-olm-master-e2e-gcp/1416088009018707968#1:build-log.txt%3A5297

awgreene: I have set the priority class to `openshift-user-critical`, which will allow the job to be preempted by user workloads and OOMKilled.

manifests/0000_50_olm_07-collect-profiles.cronjob.yaml
awgreene: Added in the latest commit.

operator-lifecycle-manager.Dockerfile
awgreene: I was too. Now that the PR is in a working state I will attempt to remove it and diagnose why the binary could not be ran.

Makefile
```diff
@@ -23,11 +23,12 @@ API_PKG := $(GO_PKG)/api
...
 
 PSM := $(addprefix bin/, psm)
+COLLECT_PROFILES_CMD := $(addprefix bin/, collect-profiles)
 OPM := $(addprefix bin/, opm)
 OLM_CMDS  := $(shell go list -mod=vendor $(OLM_PKG)/cmd/...)
-PSM_CMDS := $(shell go list -mod=vendor github.com/openshift/operator-framework-olm/cmd/...)
+PSM_CMDS := $(shell go list -mod=vendor github.com/openshift/operator-framework-olm/cmd/package-server-manager)
 REGISTRY_CMDS  := $(addprefix bin/, $(shell ls staging/operator-registry/cmd | grep -v opm))
-
+OPENSHIFT_CMDS := $(shell go list -mod=vendor github.com/openshift/operator-framework-olm/cmd/collect-profiles)
```
timflannagan, line 31: I think we may want to go in a different direction here and generalize this further. How do we feel about consolidating both of these downstream binaries using something like `ROOT_CMDS` and building using a single target?

The `build/olm` target would remove things like `$(PSM_CMDS)` and `$(OPENSHIFT_CMDS)` and instead just specify `$(MAKE) $(ROOT_CMDS) $(OLM_CMDS)`. Note: there may be a better name for this than root cmd.

awgreene: I agree with you in principle, but the command we've generated should be removed when OpenShift defines its continuous profiling strategy in the future. Given our goals to deprecate this command, could we leave it as its own binary while committing to moving the package-server command to a subcommand of a new binary (in a separate PR)?

manifests/0000_50_olm_00-pprof-rbac.yaml
```diff
@@ -0,0 +1,44 @@
...
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+  name: collect-profiles
+  namespace: openshift-operator-lifecycle-manager
+rules:
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["get", "list", "create", "delete"]
+  - apiGroups: [""]
+    resources: ["secrets"]
```
benluddy, line 15:
```suggestion
    resources: ["secrets"]
    resourceNames: ["pprof-cert"]
```

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,339 @@
...
+	configPath   string
+	clientCAPath string
+)
+
+func init() {
+	rootCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", "default", "The Kubernetes namespace where the generated configMaps should exist. Defaults to \"default\".")
+	rootCmd.MarkFlagRequired("namespace")
+	rootCmd.PersistentFlags().StringVarP(&configPath, "config-path", "c", "/etc/config", "The path to the collect-profiles configuration file.")
+	rootCmd.MarkFlagRequired("config-file")
+	rootCmd.PersistentFlags().StringVarP(&clientCAPath, "client-ca", "", "/var/run/secrets/serving-cert", "The path to the tls cert used by the client making https requests against the pprof URLs.")
```
benluddy, line 49: It might improve clarity to rename this (maybe to just `cert`?), since it's a path to the directory containing the certificate & key and not to a CA bundle.

awgreene: Renamed the following flags:
- `config-path` to `config-mount-path`
-  `client-ca` to `cert-mount-path`

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+
+func getTruePointer() *bool {
+	trueBool := true
+	return &trueBool
+}
+
+func newCmd() *cobra.Command {
+	var cfg config.Configuration
+	return &cobra.Command{
+		Use:   "collect-profiles configMapName:url",
```
timflannagan, line 72: It looks like we already have a relatively fleshed out set of flags. Any reason the configmapName and URL arguments cannot also be flags, or is this kind of setup simply preference?

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+		Short: "Retrieves the pprof data from a UNL and stores it in a configMap.",
```
timflannagan, line 73: s/UNL/URL

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+		Long: `The collect-profiles command makes https requests against pprof URLs
+		provided as arguments and stores that information in immutable configMaps.`,
+		Version: "0.0.1",
```
timflannagan, line 76: I don't see any precedent for versioning OLM binaries using this flag. Any reason we would want to include this flag going forward?

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
```
timflannagan, line 26: Would be possible to avoid vendoring another logging client and use one of the existing ones? IIRC logrus should at least be available.

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+		},
+	}
+}
+
+func verifyCertAndKeyExist(certPath, keyPath string) error {
+	fi, err := os.Stat(certPath)
+	if err != nil {
+		return err
+	}
+	if fi.Size() == 0 {
```
timflannagan, line 208: Trying to wrap my head around when a file would exist but contain zero bytes. A `$ touch <file>` kind of operation?

pkg/profiling/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+	Client     client.Client
+	Scheme     *runtime.Scheme
+}
+
+func (c *Configuration) Load() error {
+	// creates the in-cluster config
+	restConfig, err := rest.InClusterConfig()
+	if err != nil {
+		return err
+	}
```
timflannagan, line 26: Any hesitance towards always relying on the in-cluster config? The only immediate concern would be attempting to develop locally and running the binary from a remote host, which should error/produce a panic if it cannot load the in-cluster configuration. The latter can be implemented as a follow-up later down the line, but trying to wrap my head around whether there are similar situations to look out for here.

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+	var cfg config.Configuration
+	return &cobra.Command{
+		Use:   "collect-profiles configMapName:url",
+		Short: "Retrieves the pprof data from a UNL and stores it in a configMap.",
+		Long: `The collect-profiles command makes https requests against pprof URLs
+		provided as arguments and stores that information in immutable configMaps.`,
+		Version: "0.0.1",
+		PersistentPreRunE: func(*cobra.Command, []string) error {
+			return cfg.Load()
+		},
```
timflannagan, line 79: Had a quick question as to why we wouldn't be able to bake this into the RunE handler. Affects startup performance in the latter case?

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+			}
+
+			// Update serving cert after a successful run
+			return populateServingCert(cmd.Context(), cfg.Client)
+		},
+	}
+}
+
+func verifyCertAndKeyExist(certPath, keyPath string) error {
+	fi, err := os.Stat(certPath)
```
timflannagan, line 204: Any idea on whether os.Stat opens the file as well? It wasn't immediately clear to me what syscall was being used under-the-hood, and whether we need to explicitly close the file descriptor as well.

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+			return cfg.Load()
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			if len(args) == 0 {
+				klog.Info("No arguments provided, exiting")
+				return nil
+			}
+
+			jobConfig, err := config.GetConfig(configMountPath)
```
timflannagan, line 87: What happens when the configMountPath variable (or even the default /etc/config location) doesn't exist? It seems like it would produce an error attempting to open that file when reading the code in the config package. Is that the expected behavior we'd like, or would we rather handle that error and attempt to create that path for them?

timflannagan (commented):
Nice, these changes look pretty far along. I'll take another pass later.

I had a couple of comments/questions/etc. before I had to context switch to something else.

I'll play around with this work locally later.

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+func populateServingCert(ctx context.Context, client client.Client) error {
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "pprof-serving-secret",
+			Namespace: "openshift-operator-lifecycle-manager",
+		},
+	}
+	err := client.Get(ctx, types.NamespacedName{Namespace: "openshift-operator-lifecycle-manager", Name: "pprof-cert"}, secret)
```
timflannagan, line 314: Would it make sense for either of these namespace or secret name values to be configurable that default to the OLM and "pprof-secret" values? If not, these values could be live as constant variables in this package.

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+				return fmt.Errorf("error deleting expired pprof configMaps: %v", errs)
+			}
+
+			certPath := filepath.Join(certMountPath, corev1.TLSCertKey)
+			keyPath := filepath.Join(certMountPath, corev1.TLSPrivateKeyKey)
+
+			if err := verifyCertAndKeyExist(certPath, keyPath); err != nil {
+				klog.Infof("error verifying provided cert and key: %v", err)
+				klog.Info("generating a new cert and key")
+				return populateServingCert(cmd.Context(), cfg.Client)
```
timflannagan, line 138: Should we be returning here, or just attempting to create/update/etc. this serving cert secret? What happens if this cert/key path doesn't exist? Do we run this binary once, those files are created, and we run this binary again?

cmd/collect-profiles/main.go
timflannagan: Ah, I realize this is because we can provide a series of arguments. Might be still worth fleshing out the long description then at some point.

awgreene: >Ah, I realize this is because we can provide a series of arguments. Might be still worth fleshing out the long description then at some point.

Agreed.

cmd/collect-profiles/main.go
awgreene: Let's drop it then!

cmd/collect-profiles/main.go
awgreene: It's a bit of a chicken and an egg problem.

The service-ca operator cannot generate certs for use with client authentication. We elected to have the CronJob generate a self-signed cert and rotate the cert with each run. The tls secret containing the self-signed cert must be mounted to the pod, so it has to exist before the pod can populate it. You cannot omit the `tls.key` or `tls.crt` values from the data of a TLS Secret, so they had to be set to "". The initial run of the CronJob will have two empty files names `tls.key` and `tls.crt`.

The entire point of the above is to handle that initial run, update the secret, exit gracefully, OLM/Catlog Operator Deployments will then notice the updated secret and use that for the client, and allow subsequent CronJob runs to use that secret to store pprof data.

pkg/profiling/config/config.go
awgreene: It seems reasonable if people decide to use this binary outside of containers, but I'd like to see this work tacked on post code freeze.

cmd/collect-profiles/main.go
awgreene: Great question and something worth looking up! Based on the answer to [this Stack Overflow question](https://stackoverflow.com/questions/50311374/does-golangs-os-stat-open-the-file), os.Stat is a cross-platform abstraction for the Unix call stat which allows one to retrieve a metadata file's metadata without having to open it.

So no, one does not need to close the file.

cmd/collect-profiles/main.go
```diff
@@ -0,0 +1,364 @@
...
+}
+
+func getTruePointer() *bool {
+	trueBool := true
+	return &trueBool
+}
+
+func newCmd() *cobra.Command {
+	var cfg config.Configuration
+	return &cobra.Command{
```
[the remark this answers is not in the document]
awgreene: This is a great idea!

cmd/collect-profiles/main.go
awgreene: I simply thought that retrieving the KUBECONFIG outside of the command logic made more sense.

cmd/collect-profiles/main.go
awgreene: Sure

awgreene: I like klog though :(

cmd/collect-profiles/main.go
awgreene: Let's go with constants given that no other project will probably ever use this binary.

pkg/profiling/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+const (
+	disabled = "disabled"
+)
+
+type config struct {
+	Disabled bool
+}
+
+func GetConfig(path string) (*config, error) {
+	file, err := os.Open(filepath.Join(path, "disabled"))
```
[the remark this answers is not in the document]
awgreene: Sure

cmd/collect-profiles/main.go
timflannagan: On a recheck of these changes, it looks like klog is already present, and it's only being marked as an explicit dependency in the go.sum now. I don't have a strong opinion and was namely interested in reducing the vendor surface area, so it should be fine to continue using this package.

cmd/collect-profiles/main.go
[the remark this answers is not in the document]
timflannagan: Err, sorry I commented the wrong option - I might the `SilenceUsage: true` field.

cmd/collect-profiles/main.go
awgreene: >Should we be returning here, or just attempting to create/update/etc. 

We should return here. The CronJob runs every 15 minutes. Mounted secrets are only refreshed every 15 minutes. Even if we update and create these files, the OLM and Catalog pods will take at least a minute to notice and respond to the change.

awgreene: >What happens if this cert/key path doesn't exist?

If the secret doesn't exist an error is returned.
If the files are empty (which happens on the first run) this method is called earlier and initializes the values.

>Do we run this binary once, those files are created, and we run this binary again?

Correct. First run is just to initialize the secret. This happens once on a cluster. Subsequent runs (even on upgrades) do not require initialization.

cmd/collect-profiles/main.go
awgreene: So when shipped with openshift and using the manifests generated by `make manifests`, CVO will always ensure that the default jobConfig configMap exists. If a user chose to delete this map, it would normalize with a subsequent run.

awgreene: It seems reasonable to return a default jobConfig if the file isn't found.

awgreene: On second thought, it seems better for the job to exit if a user provides an invalid configuration. 

Consider the only configuration available to users, `disabled = true`. When this field is set to true, the container "completes" and exits gracefully. If a user "disabled" the job incorrectly (ie, `false` instead of `False`), by defaulting to values the container would still exit gracefully and the user may not know that the job was not disabled. Alternatively, the existing implementation raises an container Error.

pkg/profiling/config/config.go
```diff
@@ -2,10 +2,9 @@ package config
 
 import (
 	"io/ioutil"
-	"os"
 	"path/filepath"
-	"strings"
 
+	"gopkg.in/yaml.v2"
```
timflannagan, line 7: Just something to call out - you _may_ run into weird serialization errors using this yaml package, namely around handling of default values which I ran into recently between the sno/rukpak work. Might be worth using `k8s.io/apimachinery/pkg/util/yaml` instead as I've had more reliable results using that package.

awgreene:
Waiting on #129 
/hold

cmd/collect-profiles/main.go
awgreene: Addressed in the latest commit.

timflannagan:
Removing the hold as the downstream port has landed.

/hold cancel

lilic:
Hey 👋 @awgreene since there is not much description here, I am curious if you performed any performance impact on the cluster, as continuous profiling and storing to configmaps to me seems very expensive. Or if there is a proposal for this somewhere where I can read more about this and the impact it will have? Thanks!

awgreene:
Hello @lilic - It should not be expensive as the solution generates at a maximum 3 configMaps worth of data, which has a hard limit set by the object store (I believe it's 2 mb per configMap so a maximum of 6mb).

Continuous profiling was implemented with a cronjob which essentially does three things:
1. The job attempts to clean up all configMaps it has created (identified with a label) that are not the most recent versions. The job will attempt to delete each configMap and if any errors were encountered it will exit before creating any new configMaps. This approach prevents the job from creating new configMaps when it has issues deleting old configMaps. It also leaves the most recent configMap available if it has any issues creating a new configMap in step 2.
2. The job then gathers and stores the new profiling data in a configMap.
3. Delete the previous most recent configMap identified in step 1.

awgreene:
Some additional context @lilic 

Historically the OLM and Catalog operators haven't  had profiling data available when an issue was submitted.  Profiling can only be enabled by:
- Marking operator as unmanaged by CVO
- Modifying the operator's deployment to include a -profiling argument
 
This meant that in order to collect profiling information a customer would need to:
- Submit a ticket
- Have a member of the OLM team request profiling data and mention the steps above
- Customer performs steps above, attempts to recreate the incident, and reports back to us
- OLM checks if customer managed to capture the issue in profling data

In an effort to improve that workflow, OLM now enables profiling by default and saves the latest successful scrape. ConfigMaps are automatically collected by must-gather.

PR #117, opened 2021-07-14
[release-4.8] Bug 1982252: remove cleanup from startup

kevinrizza:
a series of cleanup steps run on start, but they are only required
when updating from very old versions of olm. it doesn't make sense to
carry them forever.

Upstream-commit: fc28c1f2b345f7ea4ebacf99f4533ac9b50face8


timflannagan:
I didn't see any vendor changes so this'll probably fail the verify check.

vendor/github.com/operator-framework/operator-lifecycle-manager/cmd/olm/cleanup.go
```diff
@@ -1,292 +0,0 @@
-package main
```
njhale, line 1: We should factor extra-staging changes into their own commits to avoid potential conflicts later.

PR #118, opened 2021-07-14
[release-4.8] Bug 1982294: Fix for resolution error on channels with deprecated inner entries.

kevinrizza:
Resolution would fail with an error in the presence of channels
    containing more than two entries with a deprecated inner
    entry (e.g. V1 -> V2 -> V3, with deprecated V2). This was due to
    special handling that filtered out deprecated bundles at the cache
    query layer. A channel's complete replaces-chain must be returned from
    cache queries in order to correctly determine the relative order
    between entries in the channel.

    Signed-off-by: Ben Luddy <bluddy@redhat.com>

    Upstream-commit: 5afaa1d65ec6b451cf659a09956c6f3173dabf9c


sdodson:
Re-validating bugzilla status after prow config updates.
/bugzilla refresh

PR #120, opened 2021-07-14
Bug 1979525: OLM blocks minor OpenShift upgrades when incompatible optional operators are installed

njhale:
## Description

Backports [OLM-2074](https://issues.redhat.com/browse/OLM-2074) to 4.8.z -- OLM blocks minor OpenShift upgrades when incompatible optional operators are installed.

This patch includes commits from:

- https://github.com/operator-framework/operator-lifecycle-manager/pull/2081
- https://github.com/operator-framework/operator-lifecycle-manager/pull/2103
- https://github.com/operator-framework/operator-lifecycle-manager/pull/2105
- https://github.com/operator-framework/operator-lifecycle-manager/pull/2270

## Motivation

There will be a breaking API change introduced in 4.9.0: v1beta1 CRDs are removed. Many optional operators deployed by OLM still use v1beta1 CRDs.

In order to prevent an upgrade from 4.8 to 4.9 that results in breaking these operators, at minimum we need:
- A way for operator authors to signify which minor versions of OpenShift their operator is compatible with
- OLM to signal to CVO – the thing that controls cluster upgrades – when operators, incompatible with the next minor version of OpenShift, are installed on a cluster

This patch enables the **second** bullet point above.

The story for this (https://issues.redhat.com/browse/OLM-2074) missed 4.8 feature-freeze, so this must be introduced as a bug fix to 4.8.z.


njhale:
needs to also include https://github.com/operator-framework/operator-lifecycle-manager/pull/2270 after it downstreams

/hold

njhale:
fix added

/hold cancel

njhale:
cc @ecordell @bparees for patch manager approval purposes

bparees:
Mrunal is patch manager this week, but you need to resolve the "invalid bug" label before this PR will be on his radar.

njhale:
@bparees yup -- the 4.9.0 BZ is in QE right now and I'll switch this over to 4.8.z as soon as the target release is available in bugzilla.

PR #121, opened 2021-07-14
Update the layout and content within the root scripts directory

timflannagan:
Update the README.md and remove the outdated downstream port guide command(s). Instead, reference the google doc that outlines the current process as we iterate on a more concrete process that will eventually live downstream. This is mainly a stop gap to avoid the OCP release cadence where we can't update something like the scripts/README.md during periods inside of feature/code freeze without a valid bug.

Remove the unused bash scripts. This mainly includes the helper scripts that attempted to automate the deliverance of upstream commits to the staging repositories as the presence of these scripts were generating confusion on the process. Remove the nested scripts/olm directory as those also appear to be unused.

Update the scripts/*.patch.yaml YQ patch files that are used to update the rendered staging OLM helm charts to make the requisite downstream-specific modifications to their own directory. Those patch files now live in scripts/patches and the manifest generation script has been updated to point to that new location.


timflannagan:
cc @kevinrizza - IIRC, we had talked about removing the pull/push remote scripts that were causing confusion during the last team retro and updating the README.md to simply point to the google doc that's currently housing the downstream guide.

PR #124, opened 2021-07-20
(wip) Sync with upstream master

kevinrizza:


manifests/0000_50_olm_02-services.yaml
```diff
@@ -14,9 +14,9 @@ spec:
   type: ClusterIP
   ports:
     - name: https-metrics
-      port: 8081
+      port: metrics
       protocol: TCP
-      targetPort: metrics
+      targetPort: 8080
```
timflannagan, line 19: @benluddy Are these values backwards? It looks like the service type expects the port to be an integer:

```yaml
$ k explain svc.spec.ports
...
   port	<integer> -required-
     The port that will be exposed by this service.
   targetPort	<string>
     Number or name of the port to access on the pods targeted by the service.
     Number must be in the range 1 to 65535. Name must be an IANA_SVC_NAME. If
     this is a string, it will be looked up as a named port in the target Pod's
     container ports. If this is not specified, the value of the 'port' field is
     used (an identity map). This field is ignored for services with
     clusterIP=None, and should be omitted or set equal to the 'port' field.
     More info:
     https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service
...
```

kevinrizza:
updating to remove some of the tests, bisecting the commit set

manifests/0000_50_olm_02-services.yaml
awgreene: These should be flipped. https://github.com/operator-framework/operator-lifecycle-manager/pull/2273

awgreene: This should actually be fixed by changing internalPort and externalPort to 8443 in the the values.yaml file.

PR #130, opened 2021-07-22
BUG 1984538: Remove monitoring label from openshift-operators namespace

awgreene:
This commit updates the generate_crd_manifests.sh script
so it does not enable cluster monitoring on the openshift-operators
namespace.


timflannagan:
We may want to get another downstream port before landing anything out of band.

/lgtm
/hold

PR #136, opened 2021-07-27
Bug 1985697: Update the package-server-manager leader election configuration

timflannagan:
Update the package-server-manager controller and update the leader
election intervals. This is in order to comply with the need for OCP
components to be able to withstand 60s of API server disruption on
SNO-enabled clusters.

For more information, see the following resources:
- https://github.com/openshift/library-go/pull/1104#discussion_r649313822
- https://bugzilla.redhat.com/show_bug.cgi?id=1985697

Alternative implementations include disabling leader election entirely
for SNO-enabled clusters. This implementation is centered around
dynamically querying for the Infrastructure/cluster singleton resource,
checking the HA/non-HA expectations being exposed, and setting leader
election properly. This implementation would still need to be careful
about how to handle transient errors and provide an escape hatch (e.g.
prefer enablement of leader election through a CLI flag, vs. the dynamic
value) that users can pass to the PSM deployment for failed upgrades.

### DIFF
--- a/cmd/package-server-manager/main.go
+++ b/cmd/package-server-manager/main.go
@@ -6,13 +6,9 @@ import (
 
 	"github.com/spf13/cobra"
 
-	configv1 "github.com/openshift/api/config/v1"
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 
 	"k8s.io/apimachinery/pkg/fields"
-	"k8s.io/apimachinery/pkg/runtime"
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


cmd/package-server-manager/main.go
```diff
@@ -71,6 +67,9 @@ func run(cmd *cobra.Command, args []string) error {
 		LeaderElection:          !disableLeaderElection,
 		LeaderElectionNamespace: namespace,
 		LeaderElectionID:        leaderElectionConfigmapName,
+		RetryPeriod:             timeDurationPtr(defaultRetryPeriod),
+		RenewDeadline:           timeDurationPtr(defaultRenewDeadline),
+		LeaseDuration:           timeDurationPtr(defaultLeaseDuration),
```
timflannagan, line 72: @kevinrizza I tracked down a different implementation after looking more into this offline. I'm concerned with the complexity needed to dynamically disable leader election on SNO-enabled clusters, while accounting for API server disruptions in HA clusters, and having to provide a potential escape hatch. After diving into the BZ and the linked PR, it should be fine to use the recommended leader election defaults that were provided. Any concerns with this approach? It seems like the most straight-forward implementation.

kevinrizza: sgtm

PR #138, opened 2021-07-28
chore(ci): add verify-commits make target

njhale:
Add the `verify-commits` Make target to verify that commits added between the current
head and master:
- reference only one upstream, if any
- if referencing an upstream, only make changes to the vendor,
  manifests, and/or respective staging directory
  
Also, include `verify-commits` under the `verify` target. This ensures the verification is executed by Prow w/o requiring any updates to the Prow config.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,10 @@ verify-manifests: manifests
 verify-nested-vendor:
 	./scripts/check-staging-vendor.sh
 
+.PHONY: verify-commits
+verify-commits:
+	./scripts/verify_commits.sh $(PULL_BASE_SHA) # see https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables
+
 .PHONY: verify
 verify:
 	echo "Checking for unstaged root vendor changes"
@@ -155,6 +159,8 @@ verify:
 	$(MAKE) verify-manifests
 	echo "Checking for unsynced nested [go.mod,go.sum] files"
 	$(MAKE) verify-nested-vendor
+	echo "Checking commit integrity"
+	$(MAKE) verify-commits
 
 .PHONY: help
 help: ## Display this help.
--- a/scripts/verify_commits.sh
+++ b/scripts/verify_commits.sh
@@ -1,0 +1,140 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+function err() {
+    echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $*" >&2
+}
+
+function info() {
+    echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $*"
+}
+
+function verify_staging_sync() {
+    local remote="${1}"
+    local downstream_commit="${2}"
+    local staging_dir="staging/${remote}"
+
+    local outside_staging
+    outside_staging="$(git show --name-only "${downstream_commit}" -- ":!${staging_dir}" ':!vendor :!manifests')"
+    if [[ -n "${outside_staging}" ]]; then
+        err "downstream staging commit ${downstream_commit} changes files outside of ${staging_dir}, vendor, and manifests directories"
+        err "${outside_staging}"
+        err "hint: factor out changes to these directories into a standalone commit"
+        return 1
+    fi
+}
+
+function verify_downstream_only() {
+    local downstream_commit="${1}"
+
+    local inside_staging
+    inside_staging="$(git show --name-only "${downstream_commit}" -- staging)"
+    if [[ -n "${inside_staging}" ]]; then
+        err "downstream non-staging commit ${downstream_commit} changes staging"
+        err "${inside_staging}"
+        err "only staging commits (i.e. from an upstream cherry-pick) may change staging"
+
+        return 1
+    fi
+}
+
+function upstream_ref() {
+    local downstream_commit="${1}"
+
+    local log
+    log="$(git log -n 1 "${downstream_commit}")"
+
+    local -a upstream_repos
+    mapfile -t upstream_repos < <(echo "${log}" | grep 'Upstream-repository:' | awk '{print $2}')
+
+    local -a upstream_commits
+    mapfile -t upstream_commits < <(echo "${log}" | grep 'Upstream-commit:' | awk '{print $2}')
+
+    if (( ${#upstream_repos[@]} < 1 && ${#upstream_commits[@]} < 1 )); then
+        # no upstream commit referenced
+        return 0
+    fi
+
+    
+    local invalid
+    if (( ${#upstream_repos[@]} != 1 )); then
+        err "downstream staging commit ${downstream_commit} references an invalid number of repos:"
+        err "${upstream_repos[@]}"
+        err "staging commits must reference a single upstream repo"
+        invalid=true
+    fi
+
+    if (( ${#upstream_commits[@]} != 1 )); then
+        err "downstream staging commit ${downstream_commit} references an invalid number of upstream commits:"
+        err "${upstream_commits[@]}"
+        err "staging commits must reference a single upstream commit"
+        invalid=true
+    fi
+
+    if [[ "${invalid}" == true ]]; then
+        return 1
+    fi
+
+    if git branch -r --contains "${upstream_commits[0]}" | grep -vq "${upstream_repos[0]}/.*"; then
+        err "downstream staging commit ${downstream_commit} references upstream commit ${upstream_commits[0]} that doesn't exist in ${upstream_repos[0]}"
+        err "staging commits must reference a repository containing the given upstream commit"
+        return 1
+    fi
+
+    echo "${upstream_repos[0]}"
+    echo "${upstream_commits[0]}"
+}
+
+function fetch_remotes() {
+    local -a remotes
+    remotes=(
+        'api'
+        'operator-registry'
+        'operator-lifecycle-manager'
+    )
+    for r in "${remotes[@]}"; do
+        local url
+        url="https://github.com/operator-framework/${r}.git"
+
+        git remote add "${r}" "${url}" 2>/dev/null || git remote set-url "${r}" "${url}"
+        git fetch -q "${r}" master
+    done
+}
+
+function main() {
+    fetch_remotes || { err "failed to fetch remotes" &&  exit 1; }
+
+    local target_branch="${1:-master}"
+
+    # get all commits we're introducing into the target branch
+    local -a new_commits
+    mapfile -t new_commits < <(git rev-list --no-merges HEAD "^${target_branch}")
+    info "detected ${#new_commits[@]} commit(s) to verify"
+
+    local -a sr
+    local short
+    for commit in "${new_commits[@]}"; do
+        short="${commit:0:7}"
+        info "verifying ${short}..."
+        info "$(git log -n 1 "${commit}")"
+
+        mapfile -t sr < <(upstream_ref "${commit}" || exit 1)
+        if (( ${#sr[@]} < 2 )); then # the ref contains a tuple if the values were properly parsed from the commit message
+            # couldn't find upstream cherry-pick reference in the commit message
+            # assume it's downstream-only commit
+            info "${short} does not reference upstream commit, verifying as downstream-only"
+            verify_downstream_only "${commit}" || exit 1
+        else
+            # found upstream cherry-pick reference in commit message
+            # verify as an upstream sync
+            info "${short} references upstream commit (${sr[*]}), verifying as upstream staging sync"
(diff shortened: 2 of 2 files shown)


scripts/verify_commits
```diff
@@ -0,0 +1,133 @@
...
+}
+
+function verify_staging_diff() {
+    local remote="${1}"
+    local upstream_commit="${2}"
+    local downstream_commit="${3}"
+    local staging_dir="staging/${remote}"
+
+    local outside_staging
+    outside_staging="$(git show --name-only "${downstream_commit}" -- ":!${staging_dir}")"
```
njhale, line 14: should this be "outside of the staging dir _and_ vendor"?

scripts/verify_commits
```diff
@@ -0,0 +1,133 @@
...
+    fi
+
+    local divergence
+    divergence="$(git diff --stat "${upstream_commit}" "${downstream_commit}:${staging_dir}" -- ':!vendor')"
+    if [[ -n "${divergence}" ]]; then
+        err "downstream staging commit ${downstream_commit} diverges from upstream commit ${upstream_commit}:"
+        err "${divergence}"
+        err "changes must match referenced upstream commit ${upstream_commit}"
+        return 1
+    fi
```
njhale, line 29: should I remove this? it might cause problems if we need to make some changes to accommodate diffs in staging due to any downstream-only changes to staging.

scripts/verify_commits
```diff
@@ -0,0 +1,133 @@
...
+
+    local inside_staging
+    inside_staging="$(git show --name-only "${downstream_commit}" -- staging)"
+    if [[ -n "${inside_staging}" ]]; then
+        err "downstream non-staging commit ${downstream_commit} changes staging"
+        err "${inside_staging}"
+        err "only staging commits (i.e. from an upstream cherry-pick) may change staging"
+
+        return 1
+    fi
```
njhale, line 43: we should be able to make downstream-only changes to staging, so should I change this to check for something else? Maybe "doesn't make changes to more than one staging directory at once"?

scripts/verify_commits
```diff
@@ -0,0 +1,133 @@
...
+        if git remote show "${r}"; then
+            git remote set-url "${r}" "${url}"
+        else
+            echo "remote ${r} doesn't exist"
+            git remote add "${r}" "${url}"
+        fi
+
+        git fetch "${r}" master
+    done
+}
```
njhale, line 106: there's a preexisting function for this hidden away, but it depends on the tracking file, which I don't think we need going forward. should I drop that and clean up the unused scripts in this PR too?

Makefile
```diff
@@ -155,6 +160,8 @@ verify:
 	$(MAKE) verify-manifests
 	echo "Checking for unsynced nested [go.mod,go.sum] files"
 	$(MAKE) verify-nested-vendor
+	echo "Checking commit integrity"
+	$(MAKE) verify-commits
```
njhale, line 164: just following the convention here, but is there a reason we don't just tack these on to `verify` as prereqs?

Makefile
```diff
@@ -147,6 +147,11 @@ verify-manifests: manifests
 verify-nested-vendor:
 	./scripts/check-staging-vendor.sh
 
+.PHONY: verify-commits
+verify-commits:
+	echo "$$(env)"
```
njhale, line 152: this is going to be removed, I just want to see what variables we have to work with wrt in openshift-ci

scripts/verify_commits
kevinrizza: won't this explicitly break what we're doing now with manifest changes? we are amending commits to do that right now so the changes are as idempotent as possible

at the very least, we need to ignore vendor changes as well

Makefile
timflannagan: https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables is typically a good overview of what environment variables are exposed to test steps.

scripts/verify_commits.sh
```diff
@@ -0,0 +1,117 @@
...
+        'operator-lifecycle-manager'
+    )
+    for r in "${remotes[@]}"; do
+        local url
+        url="https://github.com/operator-framework/${r}.git"
+
+        git remote add "${r}" "${url}" || git remote set-url "${r}" "${url}"
+        git fetch "${r}" master
+    done
+}
```
timflannagan, line 85: There's probably value in extracting this to a common/helper library such that people can source this in the future, but that can be done later down the line.

scripts/verify_commits.sh
```diff
@@ -0,0 +1,117 @@
...
+    remotes=(
+        'api'
+        'operator-registry'
+        'operator-lifecycle-manager'
+    )
+    for r in "${remotes[@]}"; do
+        local url
+        url="https://github.com/operator-framework/${r}.git"
+
+        git remote add "${r}" "${url}" || git remote set-url "${r}" "${url}"
```
timflannagan, line 82: Nothing pressing - IIRC, `git remote add ...` will produce at least an error message if the remote already exists, which we may want to flush.

timflannagan: I guess we could also skip the `git remote add` call by parsing `git remote -v`?

scripts/verify_commits.sh
```diff
@@ -0,0 +1,117 @@
...
+        if (( ${#sr[@]} < 2 )); then # the ref contains a tuple if the values were properly parsed from the commit message
+            # couldn't find upstream cherry-pick reference in the commit message
+            # assume it's downstream-only commit
+            info "${commit:0:7} does not reference upstream commit, verifying as downstream-only"
+        else
+            # found upstream cherry-pick reference in commit message
+            # verify as an upstream sync
+            info "${commit:0:7} references upstream commit (${sr[*]}), verifying as upstream sync"
+            verify_upstream_sync "${sr[0]}" "${commit}" || exit 1
+        fi
```
timflannagan, line 112: Hmm it's possible I'm misreading this but this seems like it's avoiding the edge case where we're pulling in an upstream commit, but someone forgot to populate that upstream commit metadata into the pulled down commit?

timflannagan: Ah, yeah I glossed over the commented text - it seems like we'd want to enforce this behavior as a stop gap until automation has been introduced. Would it be more reasonable to assume that anything that touches the staging tree _needs_ to be treated as an upstream commit? I still need to dive  into that upstream_ref function so I may be missing some context here.

scripts/verify_commits.sh
```diff
@@ -0,0 +1,117 @@
...
+    local downstream_commit="${1}"
+
+    local log
+    log="$(git log -n 1 "${downstream_commit}")"
+
+    local -a upstream_repos
+    mapfile -t upstream_repos < <(echo "${log}" | grep 'Upstream-repository:' | awk '{print $2}')
+
+    local -a upstream_commits
+    mapfile -t upstream_commits < <(echo "${log}" | grep 'Upstream-commit:' | awk '{print $2}')
```
timflannagan, line 40: Can we also add validation logic such that we ensure that both of these fields are present vs. one is present and another is not?

scripts/verify_commits.sh
```diff
@@ -0,0 +1,117 @@
...
+
+    if (( ${#upstream_repos[@]} < 1 && ${#upstream_commits[@]} < 1 )); then
+        # no upstream commit referenced
+        return 0
```
timflannagan, line 44: I think we'll need to avoid returning here, and check whether this commit touches any of the staging trees?

njhale: This return indicates a downstream-only change and if we have further downstream validations, I think they should be run outside of this function depending on the return value.

scripts/verify_commits.sh
njhale: I was parsing the remote earlier but was having some trouble with certain git command side effects. Decided this would be simpler.

scripts/verify_commits
timflannagan: I'd be down to clean up any unused scripts here - I have a PR open that attempts to cleanup various things in the repository: https://github.com/openshift/operator-framework-olm/pull/121/. Might be worth cherrypicking the 18e10cfc00ca6aa83f7301c07437fda7b3240c5a commit in that PR which removes the unused bash scripts.

scripts/verify_commits.sh
timflannagan: I realize you likely wouldn't want (or need) to parse `git remote -v` output:

```bash
if ! git remote | grep "${r}" >/dev/null 2>&1; then
    git remote add ...
```

I think the current implementation is fine and the extra verbosity when running locally with existing OLM remotes is a minor detail.

scripts/verify_commits
njhale: I think that commit could make a nice PR of its own. I typically tend towards smaller, more tightly scoped PRs. WDYT?

timflannagan: Yeah that's fine with me - we can keep those changes outside of this PR.

scripts/verify_commits.sh
njhale: good call! added.

scripts/verify_commits.sh
njhale: I added some more verification elsewhere. This should be resolved.

scripts/verify_commits.sh
njhale: got it, I added some stderr redirects.

scripts/verify_commits.sh
```diff
@@ -0,0 +1,140 @@
...
+    fi
+
+    
+    local invalid
+    if (( ${#upstream_repos[@]} != 1 )); then
+        err "downstream staging commit ${downstream_commit} references an invalid number of repos:"
+        err "${upstream_repos[@]}"
+        err "staging commits must reference reference a single upstream repo"
```
kevinrizza, line 66: nit: `staging commits must reference a single upstream repo`

scripts/verify_commits.sh
```diff
@@ -0,0 +1,140 @@
...
+        err "${upstream_commits[@]}"
+        err "staging commits must reference a single upstream commit"
+        invalid=true
+    fi
+
+    if [[ "${invalid}" == true ]]; then
+        return 1
+    fi
+
+    if git branch -r --contains "${upstream_commits[0]}" | grep -vq "${upstream_repos[0]}/.*"; then
```
kevinrizza, line 81: does this cause problems if we need to revert commits upstream? I suppose in that case we need to explicitly revert the commit in order, but it could be a problem if we're on a release branch where that is not trivially possible

scripts/verify_commits.sh
```diff
@@ -0,0 +1,140 @@
...
+
+function verify_staging_sync() {
+    local remote="${1}"
+    local downstream_commit="${2}"
+    local staging_dir="staging/${remote}"
+
+    local outside_staging
+    outside_staging="$(git show --name-only "${downstream_commit}" -- ":!${staging_dir}" ':!vendor :!manifests')"
+    if [[ -n "${outside_staging}" ]]; then
+        err "downstream staging commit ${downstream_commit} changes files outside of ${staging_dir}, vendor, and manifests directories"
```
timflannagan, line 23: One quick thought: would we necessary group an upstream-specific commit and the requisite manifest changes in the same commit? Grouping the staging tree changes + any vendor changes makes sense to me, especially in the reversion case, but still trying to wrap my head around whether we'd like to enforce that manifest changes need to also live in that same commit. Any ideas?

njhale: > but still trying to wrap my head around whether we'd like to enforce that manifest changes need to also live in that same commit. Any ideas?

The current code ensures that no staging commit touches a staging directory _and_ a directory _other than_ ./vendor or ./manifests. It doesn't enforce that changes to manifests occur in a staging commit, if that's the concern.

timflannagan: Ah, my mistake likely glossed over too quickly - this implementation is fine with me.

scripts/verify_commits.sh
njhale: > does this cause problems if we need to revert commits upstream?

reverting a commit is additive -- it creates a new commit that undoes the changes of the target -- so in theory, the original commit should still exist (if I didn't misunderstand the concern).

see https://git-scm.com/docs/git-revert#_description

kevinrizza: yeah, we just need to enforce that we're not tossing history and it's not a problem. I think this is fine I just wanted to call it out as a potential problem if upstream ever got borked.

kevinrizza:
This is a script change that doesn't affect any of the running e2e tests in any way. Overriding ci 

/override ci/prow/e2e-aws-olm
/override ci/prow/e2e-gcp
/override ci/prow/e2e-upgrade

PR #140, opened 2021-07-28
[release-4.8] Bug 1986476: resolver: remove legacy support for fallback parsing of CSVs

joelanford:
When no APIs or properties are present on bundles, the resolver
currently falls back to a legacy mode where that information is parsed
from CSVs present in the index.

This legacy fallback method causes the resolver to incorrectly identify
multiple heads for a channel when:
a) The index contains CSVs only on channel head bundles, and
b) A package contains two channels, where one channel contains the head
   of the other channel and there is at least one other node between
   these channel head nodes in that channel.
c) The bundles have no properties, provided APIs, or required APIs

Conditions a) and b) are extremely prevalent, so this bug will often be
encountered whenever c) itself is true.

This commit removes support for the legacy CSV parsing fallback, which
means that only first class fields in the GRPC API will be used during
resolutions.

Upstream-commit: 7fa42934ea0ce5a5f4c589767ce88b8fac7f6c68

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/registry/resolver/operators.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/registry/resolver/operators.go
@@ -278,31 +278,6 @@ func NewOperatorFromBundle(bundle *api.Bundle, startingCSV string, sourceKey reg
 		properties = append(properties, ps...)
 	}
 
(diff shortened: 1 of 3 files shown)


timflannagan:
That failed unit test looks like it warrants further investigation.

/lgtm cancel

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 0.20

adding cherry-pick approved

PR #141, opened 2021-07-29
Add PriorityClass setting to registry pods for default CatalogSource

dinhxuanvu:
It's mandate for openshift compoments to have necessary priority class.
The registry pods that belong to default catalogsources should have
`system-cluster-critical` priority class.

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/grpc.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/grpc.go
@@ -186,17 +186,34 @@ func (c *GrpcRegistryReconciler) currentPodsWithCorrectImage(source grpcCatalogS
 	return found
 }
 
+func (c *GrpcRegistryReconciler) currentPodsWithCorrectHash(source grpcCatalogSourceDecorator, saName string) []*corev1.Pod {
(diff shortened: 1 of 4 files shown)


staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/grpc.go
```diff
@@ -186,17 +186,34 @@ func (c *GrpcRegistryReconciler) currentPodsWithCorrectImage(source grpcCatalogS
...
+	}
+	found := []*corev1.Pod{}
+	newPod := source.Pod(saName)
+	for _, p := range pods {
+		if PodHashMatch(p, newPod) {
+			found = append(found, p)
+		}
+	}
+	return found
+}
```
timflannagan, line 203: Unfortunately, and it's possible I'm missing something here, we are trying to avoid touching any code in the staging tree that hasn't been introduced/modified/etc. upstream. This is to ensure that the staging directory is 1:1 with the upstream repositories. We may need to pivot to a more generalized solution for injecting annotations into the catalogsource(s).

dinhxuanvu: I can try to convert this solution to something more upstream-friendly. However, one of the reasons we split downstream and upstream is the possibility of having downstream-only changes. This is potentially one of the cases. Where do you say I should make those changes then?

dinhxuanvu:
Is replaced by an upstream PR
/close

PR #153, opened 2021-08-03
Bug 1989398: Avoid processing the .indexignore file when walking the root fs

timflannagan:
Update internal/declcfg/load.go and avoid processing any files that are
named `.indexignore` when walking the declarative config index root
filesystem. When validating declarative config directories, the
.indexignore file was being processed and validated.

Signed-off-by: timflannagan <timflannagan@gmail.com>

Upstream-commit: 78f27b39dc098c2e56681d9a3d8e786aae95d4fe

### DIFF
--- a/staging/operator-registry/internal/declcfg/load.go
+++ b/staging/operator-registry/internal/declcfg/load.go
@@ -17,6 +17,10 @@ import (
 	"github.com/operator-framework/operator-registry/internal/property"
 )
 
+const (
+	indexIgnoreFilename = ".indexignore"
+)
+
 type WalkFunc func(path string, cfg *DeclarativeConfig, err error) error
 
 // WalkFS walks root using a gitignore-style filename matcher to skip files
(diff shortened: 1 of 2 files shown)


staging/operator-registry/internal/declcfg/load.go
```diff
@@ -36,7 +40,9 @@ func WalkFS(root fs.FS, walkFn WalkFunc) error {
 		if err != nil {
 			return walkFn(path, nil, err)
 		}
-		if info.IsDir() || matcher.Match(path, false) {
+		// avoid validating a directory, an .indexignore file, or any file that matches
+		// an ignore pattern outlined in a .indexignore file.
+		if info.IsDir() || info.Name() == indexIgnoreFilename || matcher.Match(path, false) {
```
njhale, line 45: Are there no tests for this?

njhale (approved):
Ideally, all behavioral changes should be accompanied by a test, but since this has already merged upstream, I don't think we should introduce a downstream-diff for it.

/lgtm

PR #157, opened 2021-08-04
Bug 1932139: Ensure cross-compiled opm binaries produce non-empty version information

timflannagan:
Update the root Makefile and ensure the darwin/window cross-compiled opm
binaries produce non-empty git information when running `opm version`.

Before, the first --ldflag value, which contained the git versioning
information was being overriden by the second --ldflag value.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,16 +71,16 @@ $(OLM_CMDS):
 
 $(PSM_CMD): FORCE
 	go build $(GO_BUILD_OPTS) $(GO_BUILD_TAGS) -o $(PSM_CMD) $(ROOT_PKG)/cmd/package-server-manager
-	
+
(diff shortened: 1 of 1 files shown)


timflannagan:
Going to play around with the image that gets spit out to verify those changes work outside of my local environment.

/hold

awgreene:
Looks fine provided that the binary outputs the correct version. It would be nice to include a test that prevents regressions.
/approve

timflannagan:
These changes only touch the linker options provided to the various OLM binaries.

/override ci/prow/e2e-upgrade

timflannagan:
These changes only touch the linker options provided to the various OLM binaries.

/override ci/prow/e2e-gcp
/override ci/prow/e2e-aws-olm

PR #162, opened 2021-08-11
Bug 1982781: fix(deprecate): explicit pkg rm drops latest deprecated

njhale:
- When using opm index|registry rm <pkg>, drop the "latest" related bundles
from the deprecated table. This makes "un-deprecating" a bundle
possible, which is useful if that  bundle was deprecated by
accident
- Add a migration and logic on deprecatetruncate that drops
truncated bundles (i.e. w/o assoc. channel_entry rows) from the deprecated
table, clearing the "deprecation history" so that operations post package
removal don't assume a truncated bundle should be deprecated

See https://bugzilla.redhat.com/show_bug.cgi?id=1982781 for motivation

Upstream-commit: 29e90dea6a1c2ef12075ca866284cd8f267c2bba


exdx:
/approve 
Is this a backwards-incompatible change? I guess we can just call it a bug.

njhale:
/hold

investigating CI failures

njhale:
/hold cancel

still investigating flakes, but they seem to be unrelated

PR #166, opened 2021-08-16
Bug 1993286: fix(openshift): drop z from next calculated y-stream (#2324)

njhale:
When determining operator compatibility, drop z and build
versions in the calculation of the next Y-stream (minor)
release of OpenShift.

e.g. If the current version is v4.9.5+build, the next Y-stream is
calculated as v4.10.0.

If a pre-release is included, drop it as well but don't increment
the minor version.

e.g. If the current version is v4.10.0-rc, the next Y-stream is
calculated as v4.10.0.

Before this change, the next Y-stream was the result of simply
iterating the cluster's minor version. When the cluster was at
a patch version greater than that specified by an operator, upgrades
would be erroneously blocked.

e.g. If the current version was v4.9.5, the next version would be
calculated as v4.10.5, which would block upgrades on operators with
max versions set to v4.10 -- or more explicitly [v4.10.0, v4.10.5) (')'
is read "exclusive").

Upstream-commit: 760c10d78d6f830b5aa773080847b46557c5986a

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/operators/openshift/helpers.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/operators/openshift/helpers.go
@@ -125,16 +125,20 @@ func transientErrors(err error) error {
 }
 
 func incompatibleOperators(ctx context.Context, cli client.Client) (skews, error) {
-	next, err := desiredRelease(ctx, cli)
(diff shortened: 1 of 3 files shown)


timflannagan:
/retitle Bug 1993286: fix(openshift): drop z from next calculated y-stream (#2324)

PR #181, opened 2021-09-02
Bug 2000146: Introduce olm.channel schema, remove channel metadata-related bundle properties

joelanford:
Description:
  - Define olm.channel schema to specify channel membership and upgrade edges.
    For example:

        ---
        schema: olm.channel
        package: foo
        name: stable
        entries:
          - name: foo.v0.0.1
          - name: foo.v0.0.2
          - name: foo.v0.1.0
            replaces: foo.v0.0.1
            skipRange: <0.1.0
            skips:
              - foo.v0.0.2

  - Remove properties on the olm.bundle blobs that are related to channel
    metadata. Specifically olm.channel, olm.skips, and olm.skipRange are
    removed.

  - Conversions from the model to the file-based config format use the
    olm.channel schema.

  - 'opm render' no longer populates channel-related metadata when rendering
    bundle images.

Motivation:

The current schema overlooked the fact that bundle properties contained
channel-level metadata, which goes directly against the design goals of
file-based configs.

This PR corrects that issue by implementing a new olm.channel schema and
removing the ability of opm render to pull this metadata from bundles created
with imperative sqlite-based workflows in mind.

This is also an improvement on the the existing olm.bundle properties because:

- olm.skips and olm.skipRange properties apply to all channels. This causes
  channel upgrade graphs to be defined by channel-specific replaces values from
  the olm.channel property and by the global skips and skipRange properties,
  which can make it extremely difficult to define usable upgrade graphs when
  multiple channels are involved.

- It centralizes all of the channel membership and upgrade edge information in
  a single place, making it easier for humans to visualize and for computers to
  quickly parse and generate upgrade graphs.

- It simplifies implementation of veneer APIs that focus on different upgrade
  graph definitions. Different channel and upgrade graph definitions can be built
  without touching any olm.bundle blobs.

Upstream-commit: 59d0ec7da18a467451138458b4dfafe547341b34


staging/operator-registry/internal/declcfg/declcfg_to_model_test.go
```diff
@@ -1,8 +1,10 @@
 package declcfg
 
 import (
+	"encoding/json"
 	"testing"
 
+	"github.com/operator-framework/operator-registry/internal/property"
```
ankitathomas, line 7: nit: ordering

ankitathomas (approved):

staging/operator-registry/internal/declcfg/declcfg_to_model_test.go
joelanford: I thought about fixing this here, but didn't want to cause conflicts with future cherry picks.

timflannagan:
Related: https://github.com/openshift/origin/pull/26444.

PR #193, opened 2021-09-17
Prevent incompatible operators from being resolved (injection variant)

njhale:
One of two ways to implement downstream-only constraints.

Add a downstream-only catalog operator command that injects a dependency resolution system-constraint prohibiting operators incompatible with a cluster's Y-stream release from being installed.

**Implementation Pros:**
- doesn't require carrying downstream-only patches in `/staging`
- lets us quickly inject more downstream-only constraints on-demand

**Cons:**
- depends on the constraint injection mechanism implemented in https://github.com/operator-framework/operator-lifecycle-manager/pull/2360
- more refactoring of the resolver may produce much nicer injection points


njhale:
the team decided to go with the hardcoded approach

PR #194, opened 2021-09-18
Prevent incompatible operators from being resolved (hardcode variant)

njhale:
The other way to implement downstream-only constraints (see #193 for the alternative).

Hardcode a constraint that prevents operators from being resolved on versions of OpenShift clusters they don't support.

**Implementation Pros:**
- doesn't depend on upstream refactoring
- we don't need to expose public Go APIs for an constraint injection upstream (yet)

**Cons:**
- need to carry a downstream-only patch to `/staging`
- the first downstream-only patch to `/staging` to date

**Note:** We expect `verify` to (correctly) fail since we're patching `/staging`.


njhale:
stale

PR #200, opened 2021-10-13
WIP: 4.10 downstream/upstream sync - first half

njhale:
Attempts to sync the first half of #196


timflannagan:
Upgrade failing appeared to be due to some etcd instability. The single failure in the e2e suite is a known flake upstream.

/retest

PR #204, opened 2021-10-18
4.10 downstream/upstream sync - second half

njhale:
Second half of commits in #196 

Subset of:

- #196


njhale:
superseded by #205

PR #211, opened 2021-10-26
Bug 2002276: Remove oudated subscription update logic to improve resolution delay

dinhxuanvu:
Currently, olm logic checks for upgrade in subscription via another
obsolete API that is no longer in use for dependency solution. As a
result, sometimes, subscriptions display UpgradeAvailable status but
there will be no upgrades as the upgrade is not valid in the resolver.
Also, the UpgradeAvailable status is used to trigger the new resolution
even though that status is no longer a valid indicator of having a pending
upgrade. This leads to unwanted upgrade delay when the obsolete API works
properly.

This commit will remove the code that is using this obsolete API and
allow the resolution to happen when there is a subscription change.

Signed-off-by: Vu Dinh vudinh@outlook.com


staging/operator-lifecycle-manager/pkg/controller/registry/resolver/cache/cache.go
```diff
@@ -136,9 +136,9 @@ type NamespacedOperatorCache struct {
 func (c *NamespacedOperatorCache) Error() error {
 	var errs []error
 	for key, snapshot := range c.snapshots {
-		snapshot.m.Lock()
+		snapshot.m.RLock()
 		err := snapshot.err
-		snapshot.m.Unlock()
+		snapshot.m.RUnlock()
```
jianzhangbjz, line 141: @dinhxuanvu One question, maybe I missed something, why not us use the `defer` here?

timflannagan: You wouldn't be able to use defer here as we're performing locking/unlocking operations during a for loop and the defer statement only triggers once the [function finishes execution](https://golang.org/doc/effective_go#defer). In order to use a defer statement here, you would need to wrap the lock/unlock logic into an anonymous function:

```go
for key, snapshot := range c.snapshots {
    func() {
        snapshot.m.RLock()
        defer snapshot.m.RUnlock()
        ...
    }
}
```

jianzhangbjz: @timflannagan  Thanks for your clarification! I think we should make sure the `RLock` can be released successfully. Actually, we met an issue: https://bugzilla.redhat.com/show_bug.cgi?id=2020486 that the status of the CatalogSource is `lastObservedState: IDLE` sometimes, and I checked the logs, found the warning: `level=warning msg="couldn't find service in cache"`. I guess it's related to this PR.

yuvalk:
can we backport this to 4.9?

PR #213, opened 2021-11-02
Sync w/ upstream: operator-lifecycle-manager

njhale:
Contains the set of commits in https://github.com/operator-framework/operator-lifecycle-manager/ that haven't been pulled downstream as of 11/02/2021.

When I get the chance, I'll put the scripts I used to generate this up for review as well.

Note: Some of the commits here are empty. Usually, this is due to a commit having already been cherry-picked with incorrect upstream tracking commit message trailers. I decided to include these commits as a simple way to rectify the situation -- i.e. more metadata reduces confusion in future sync attempts.


timflannagan (commented):
This is missing any of the explicit Makefile override options when running the staging OLM e2e suite in downstream environments: https://github.com/openshift/operator-framework-olm/pull/196/commits/467750e14b7745dace418a6d2495cd0f83688c37.

staging/operator-lifecycle-manager/cmd/olm/main.go
```diff
@@ -171,6 +171,11 @@ func main() {
 	op.Run(ctx)
 	<-op.Ready()
 
+	// Emit CSV metric
+	if err = op.EnsureCSVMetric(); err != nil {
+		logger.WithError(err).Fatalf("error emitting metrics for existing CSV")
+	}
```
timflannagan, line 177: This commit introduces a problematic/flaky e2e test that was prohibiting the initial downstream sync and leading to timeouts. It may be easier to simply exclude this commit until we fix that flaky e2e test upstream from this sync PR.

staging/operator-lifecycle-manager/deploy/chart/templates/0000_90_olm_01-prometheus-rule.yaml
```diff
@@ -18,6 +18,8 @@ spec:
           severity: warning
           namespace: "{{ "{{ $labels.namespace }}" }}"
         annotations:
+          summary: CSV failed for over 2 minutes
+          description: Fires whenever a CSV has been in the failed phase for more than 2 minutes.
```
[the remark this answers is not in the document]
njhale: done

staging/operator-lifecycle-manager/.github/workflows/e2e-tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '~1.16'
-      - run: make e2e-local E2E_NODES=2 JUNIT_DIRECTORY=./artifacts/
+      - run: make e2e-local E2E_NODES=2 ARTIFACTS_DIR=./artifacts/
```
timflannagan, line 17: We'll technically need to update o/release as well as we explicitly specify the `JUNIT_DIRECTORY` there. Likely easiest to drop this commit from this PR and tackle at a later date?

njhale: This was a breaking change then. It's unfortunate we didn't catch it earlier.

timflannagan: I called it out when the PR was opened: https://github.com/operator-framework/operator-lifecycle-manager/pull/2432#discussion_r741374284 - imo it's a quick change and not the end of the world as we're between freezes right now and have the flexibility to make these types of changes.

timflannagan:
Opened https://github.com/operator-framework/operator-lifecycle-manager/pull/2449 for the 'Subscription puppeting CatalogSource health status when missing target catalog should surface the missing catalog' failure.

The 'Catalog represents a store of bundles which OLM can use to install Operators adding catalog template adjusts image used' failure is known - https://github.com/operator-framework/operator-lifecycle-manager/pull/2397 took a stab at addressing that failure but it's been difficult to reproduce upstream.

The 'Operator Group static provider' failure is a known flake and already tracked upstream.

staging/operator-lifecycle-manager/cmd/olm/main.go
timflannagan: Note: I dropped this commit from the PR to help cut down on e2e flakes. We're already tracking fixing this upstream, so we're not missing a ton excluding it while we get a proper fix through the chain.

timflannagan:
(to avoid any bot spamming)

/lgtm cancel

timflannagan:
```
                TypeMeta: {Kind: "", APIVersion: ""},
                ListMeta: {
                    SelfLink: "",
                    ResourceVersion: "",
                    Continue: "",
                    RemainingItemCount: nil,
                },
                Status: "Failure",
                Message: "object is being deleted: namespaces \"test-namespace-1\" already exists",
                Reason: "AlreadyExists",
                Details: {
                    Name: "test-namespace-1",
                    Group: "",
                    Kind: "namespaces",
                    UID: "",
                    Causes: nil,
                    RetryAfterSeconds: 0,
                },
                Code: 409,
```

This should be an easy fix to avoid hardcoding the namespace's metadata.Name if it's being generated in the BeforeEach block.

/retest

timflannagan:
Both of those tests are known (and tracked upstream).

/retest

timflannagan:
Firing up an openshift cluster to try and reproduce that catalog templating e2e failure.

/retest

PR #215, opened 2021-11-17
Bug 2024048: Remove outdated subscription update logic to improve resolution delay

dinhxuanvu:
Currently, olm logic checks for upgrade in subscription via another
obsolete API that is no longer in use for dependency solution. As a
result, sometimes, subscriptions display `UpgradeAvailable` status but
there will be no upgrades as the upgrade is not valid in the resolver.
Also, the `UpgradeAvailable` status is used to trigger the new resolution
even though that status is no longer a valid indicator of having a pending
upgrade. This leads to unwanted upgrade delay when the obsolete API works
properly.

This commit will remove the code that is using this obsolete API and
allow the resolution to happen when there is a subscription change.

Upstream-commit: 81e7a60bc7a62da4a469041ce89e3867e9f47fde


dinhxuanvu:
@jianzhangbjz PTAL

jianzhangbjz:
Sorry for the late.
/lgtm

PR #220, opened 2021-12-02
[WIP] Add cluster runtime constraints handling

perdasilva:
Signed-off-by: Per G. da Silva <perdasilva@redhat.com>

# Description
This the downstream component of the work being done in https://github.com/operator-framework/operator-lifecycle-manager/pull/2498 to enable side loading of runtime constraints. Here, we add the runtime constratins file the operator image and update the olm catalog operator deployment manifest to include the environment variable (`RUNTIME_CONSTRAINTS`), which points to the runtime constraints file in the local filesystem, and enables the feature

### DIFF
--- a/runtime_constraints/README.md
+++ b/runtime_constraints/README.md
@@ -1,0 +1,11 @@
+# Cluster Runtime Constraints
+
+Cluster runtime constraints are base constraints that always get applied to the resolution process to avoid installing
+packages that might be unsuitable for the cluster (consumes too many resources, wrong kubernetes/ocp version, etc.). 
+Currently, there's no first class way to do this. Until we design the canonical way to define
+cluster runtime constraints, we are making availing a stopgap solution for IBM and OCP, we:
+  1. Are adding a file to the downstream OLM image that includes the runtime constraints
+  2. Have modified the upstream catalog operator to load the runtime constraints in to the resolver if a `RUNTIME_CONSTRAINTS` environment variable is defined
+  3. Update the deployment manifests for the olm catalog operator deployment to add the `RUNTIME_CONSTRAINTS` environment variable
+
+The upstream PR enabling this behavior is [https://github.com/operator-framework/operator-lifecycle-manager/pull/2498](#2498).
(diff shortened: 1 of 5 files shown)


runtime_constraints/README.md
```diff
@@ -0,0 +1,11 @@
+# Cluster Runtime Constraints
```
njhale, line 1: Does this belong in `docs.go`? i.e. looks like something we want in the go docs.

runtime_constraints/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Cluster runtime constraints are base constraints that always get applied to the resolution process to avoid installing
```
njhale, line 3: It seems more apt to call it something like "global" or "static" than "cluster runtime", since this could be conflated with the "virtual bundle" concept mapped out in some of the EPs.

runtime_constraints/README.md
perdasilva: Could you please be more specific for me: do you mean there's a docs place in this repo?  do you mean upstream? do you mean docstrings in the code?

My intent here is to add developer documentation - just so people get the context of why that file exists there. Once we have the _right_ way of doing things, then all of this should go away.

It is still my intention to add user facing docs though - but I was thinking of doing that in the upstream

runtime_constraints/README.md
perdasilva: damn - I see a lot of refactoring in my future now hahaha

njhale (commented):
Please see https://github.com/operator-framework/operator-lifecycle-manager/pull/2498#pullrequestreview-822924457 for my thoughts on this.

runtime_constraints/README.md
njhale: Sorry -- Go doc strings; e.g. https://github.com/kubernetes-sigs/controller-runtime/blob/master/doc.go autogenerates developer docs at https://pkg.go.dev/sigs.k8s.io/controller-runtime

PR #227, opened 2021-12-13
clean up unused scripts

ankitathomas:


timflannagan (commented):
Thanks - I think I'd like to also remove the scripts/[pull_upstream.sh,push_upstream.sh] scripts as well, just to avoid confusing anyone about the current workflow. It looks like this PR has passed most of the checks though, so I won't hold up these cleanup just for that.

We want to refactor the patch YAML files that are stored in the scripts directory and house them in their own directory as well (which would require updates to the scripts responsible for generating the root set of manifests) but that can be done another time too.

/lgtm

timflannagan:
There's no need to re-spin upgrade testing as this PR doesn't affect anything user-facing.

/override ci/prow/e2e-upgrade

PR #228, opened 2021-12-13
Bug 2048563: feat added leader election conventions

eggfoobar:
This should take advantage of leader election changes for SNO clusters proposed in this [library-go PR](https://github.com/openshift/library-go/pull/1242).

Changes:
  - updated leader election to follow convention and use different leader election for SNO topology

Signed-off-by: ehila <ehila@redhat.com>

### DIFF
--- a/pkg/leaderelection/leaderelection.go
+++ b/pkg/leaderelection/leaderelection.go
@@ -1,0 +1,77 @@
+package leaderelection
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	openshiftcorev1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/client-go/rest"
+)
+
+const (
+	infraResourceName = "cluster"
+
+	// Defaults follow conventions
+	// https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#high-availability
+	// Impl Calculations: https://github.com/openshift/library-go/commit/7e7d216ed91c3119800219c9194e5e57113d059a
+	defaultLeaseDuration = 137 * time.Second
+	defaultRenewDeadline = 107 * time.Second
+	defaultRetryPeriod   = 26 * time.Second
+)
+
+func GetLeaderElectionConfig(log logr.Logger, restClient *rest.Config, enabled bool) (defaultConfig configv1.LeaderElection) {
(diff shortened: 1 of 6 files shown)


kevinrizza:
@eggfoobar this PR seems reasonable to me, but I do want to call out that it only resolves the problem for the downstream only controller you updated, and none of the controllers that we're pulling in from the upstream operator-framework project in the staging directory.

eggfoobar:
Thanks for the heads up @kevinrizza , I think I might have misunderstood then, from what I could see from the `holderIdentity` it seemed like just the downstream operator was doing leader election, the upstream project seemed to not have that setup just yet. Is this the right spot to look for that [code](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/cmd/olm/manager.go#L50) ?

base.Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Dockerfile to bootstrap build and test in openshift-ci
 
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.10
```
perdasilva, line 3: would it make sense to split this out of the PR? Or is it a pre-requisite for the new behaviour?

eggfoobar: Sounds like a plan, reverted, it shouldn't be required by any of the changes in this PR.

pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// Defaults follow conventions
+	// https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#high-availability
+	// Impl Calculations: https://github.com/openshift/library-go/commit/7e7d216ed91c3119800219c9194e5e57113d059a
+	defaultLeaseDuration = 137 * time.Second
+	defaultRenewDeadline = 107 * time.Second
+	defaultRetryPeriod   = 26 * time.Second
+)
+
+func GetLeaderElectionConfig(log logr.Logger, restClient *rest.Config, enabled bool) (defaultConfig configv1.LeaderElection) {
```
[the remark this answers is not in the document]
eggfoobar: Sure thing, re-arranged things a bit and wrote some unit tests

eggfoobar:
Updated to include the BZ number, since this PR is related to the effort in that BZ it seemed appropriate to attach it to that.

eggfoobar:
/retitle Bug 2048563: feat added leader election conventions

eggfoobar:
Hey @dinhxuanvu seems the latest `e2e-upgrade` passed, should we remove the hold?

PR #234, opened 2021-12-16
Sync registry 12-16

anik120:


dinhxuanvu:
@anik120 This PR fails verify for some reason.

ankitathomas:
> @anik120 This PR fails verify for some reason.

The .gitignore commit doesn't have the Upstream-repository/commit trailers, either that commit needs to be taken out/amended or that check needs to be turned off

anik120:
@ankitathomas @dinhxuanvu I've dropped the `.gitignore` commit, that commit is in fact not needed in this repo. Should be good to go now.

timflannagan:
This didn't make the feature freeze cut

/hold

timflannagan:
I think we can close this out now that we're past feature freeze unless we want to market this as a bug and attach a BZ # to this PR.

/close

PR #237, opened 2021-12-16
Add the create-only annotation to the olmconfig/cluster manifest

awgreene:


scripts/generate_crds_manifests.sh
```diff
@@ -64,6 +64,7 @@ ${YQ} merge --inplace -d'0' manifests/0000_50_olm_00-namespace.yaml scripts/moni
 ${YQ} write --inplace -s scripts/olm-deployment.patch.yaml manifests/0000_50_olm_07-olm-operator.deployment.yaml
 ${YQ} write --inplace -s scripts/catalog-deployment.patch.yaml manifests/0000_50_olm_08-catalog-operator.deployment.yaml
 ${YQ} write --inplace -s scripts/packageserver-deployment.patch.yaml manifests/0000_50_olm_15-packageserver.clusterserviceversion.yaml
+${YQ} merge --inplace manifests/0000_50_olm_02-olmconfig.yaml scripts/cluster-olmconfig.patch.yaml
```
timflannagan, line 67: nit: does this need the `-s` flag like the previous patches that are listed above?

/lgtm

awgreene: The `-s` flags are included  the write commands save the changes, which merge automatically does. Additionally, merge doesn't support the `-s` flag.

PR #238, opened 2022-01-05
Bug 2037168: Remove incorrect cvo annotations

kevinrizza:
The single-node-developer profile was ultimately not implemented, and as
a result is ultimately unnecessary. This commit removes it from all
manifests
see  https://github.com/openshift/cluster-version-operator/pull/685

Additionally, the ibm-cloud-managed annotation and self-managed
annotations are applied to both versions of the psm deployment manifest.
Since these are in conflict, in the future they could potentially cause
flapping by the CVO to apply both versions of the manifest to the
cluster.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2037168

### DIFF
--- a/scripts/generate_crds_manifests.sh
+++ b/scripts/generate_crds_manifests.sh
@@ -55,7 +55,6 @@ add_ibm_managed_cloud_annotations() {
          ${YQ} d -d'*' --inplace "$g" 'spec.template.spec.nodeSelector."node-role.kubernetes.io/master"'
       fi
       ${YQ} w -d'*' --inplace --style=double "$f" 'metadata.annotations['include.release.openshift.io/self-managed-high-availability']' true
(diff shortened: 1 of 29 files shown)


scripts/generate_crds_manifests.sh
```diff
@@ -92,8 +91,6 @@ metadata:
   labels:
     app: package-server-manager
   annotations:
-    include.release.openshift.io/self-managed-high-availability: "true"
```
timflannagan, line 95: Why was this annotation removed as well?

dinhxuanvu (approved):
/lgtm

scripts/generate_crds_manifests.sh
kevinrizza: Because they're opposite profiles. The self managed profile (the default cvo profile) shouldn't apply to the manifests that are specifically just for the ibm manifests. Take a look at our other deployment manifests that specifically include them, this change just makes those match.

For the manifests where we just have one copy of the manifest, both annotations apply.

PR #240, opened 2022-01-05
Bug 2026343: Address Invalid Address in GRPC Catalogs (#2499)

awgreene:
Problem: Within the catalogSource resource, the RegistryServiceStatus
stores service information that is used to generate an address that OLM
relies on in order to establish a connection with the associated pod.
If the RegistryStatusService is not nil and is missing the namespace,
name, and port information for its service, OLM is unable to recover
until the catalogService's associated pod has an invalid image or
spec.

Solution: When reconciling a CatalogSource, OLM will now ensure that
the RegistryServiceStatus of the catalogSource is valid and will update
the catalogSource's status to reflect the change.  Additionally, this
address is stored within the status of the catalogSource within the
status.GRPCConnectionState.Address field. If the address changes, OLM
will update this field to reflect the new address as well.

Upstream-commit: cabe200bdabed1cb10c35cf434f7e426e3a8cb65

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/operators/catalog/operator.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/operators/catalog/operator.go
@@ -803,6 +803,7 @@ func (o *Operator) syncConnection(logger *logrus.Entry, in *v1alpha1.CatalogSour
 		// Set connection status and return.
 		out.Status.GRPCConnectionState.LastConnectTime = now
 		out.Status.GRPCConnectionState.LastObservedState = source.ConnectionState.String()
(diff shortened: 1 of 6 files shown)


timflannagan (commented):
It looks like you need to run `go mod vendor` locally as we should see some root vendor directory changes in this PR as well.

awgreene:
@timflannagan do I just amend that to the existing commit or do we separate that out?

timflannagan:
@awgreene Yeah I'd just amend it to the existing commit.

timflannagan:
Holding so the bot doesn't go crazy while we wait for the upgrades test to stabilize.

/hold

PR #242, opened 2022-01-17
Bug 2003164: Do not modify object from the lister cache (#2562)

awgreene:
Problem: Objects retrieved from the lister should
not be modified.

Solution: In locations where an object originating from the lister
cache must be modified, modify a copy of the object instead.

Upstream-commit: 5a7f8033dfc04150d1c21ef4e86fd7bc00bbfa39


timflannagan (approved):

timflannagan:
I make that mistake every time that approval in the organization doesn't automatically tack on a lgtm :disappointed: 

/lgtm

PR #245, opened 2022-01-24
Bug 1975543: Use CVO to delete manifests removed from OLM

awgreene:
Problem: Over the course of OLM's lifecycle resources
included in the CVO payload have been removed. CVO does
not automatically removed these resources from the cluster
unless they are properly annotated.

Solution: Reintroduce the removed manifests and include the
annotation that informs CVO to delete the resources.

### DIFF
--- a/manifests/0000_50_olm_11-olm-operators.configmap.removed.yaml
+++ b/manifests/0000_50_olm_11-olm-operators.configmap.removed.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: olm-operators
+  namespace: openshift-operator-lifecycle-manager
+  annotations:
+    release.openshift.io/delete: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
--- a/manifests/0000_50_olm_12-olm-operators.catalogsource.removed.yaml
+++ b/manifests/0000_50_olm_12-olm-operators.catalogsource.removed.yaml
@@ -1,0 +1,9 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: olm-operators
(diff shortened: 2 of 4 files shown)


awgreene:
I need to check if CVO supports nested directories in the `/manifest` dir.
/hold

manifests/removed/0000_50_olm_16-packageserver.clusterserviceversion.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: packageserver
+  namespace: openshift-operator-lifecycle-manager
+  annotations:
+    release.openshift.io/delete: "true"
```
awgreene, line 7: According to @timflannagan, the package server is still deployed as a csv downstream, so this may conflict. I am spinning up a cluster now.

timflannagan: Yeah, the package server component is still deployed using a CSV resource - we had tried to migrate off the CSV and instead deploy it using core k8s manifests, but it extended the scope of the SNO work and there was some complexity involved with migrating existing OLM releases and we didn't want to risk missing the release dates. We'll need to revisit this though, and we should ideally be tracking this $somewhere. cc @perdasilva @kevinrizza

awgreene: The packageServer CSV is actually managed by an OLM controller and is defined [here](https://github.com/openshift/operator-framework-olm/blob/3820d16f4912896c51379297e65563b3ec78d272/pkg/manifests/csv.yaml). CVO cannot be configured to delete this CSV.

dinhxuanvu (approved):
/lgtm

dinhxuanvu:
This PR and #228 basically perma-fail e2e-upgrade. Some investigation needed here to find out why.

PR #247, opened 2022-01-25
Remove benluddy from approvers.

benluddy:


kevinrizza:
/label bugzilla/valid-bug

overriding prow because this is just an owners file change

timflannagan:
Overriding - there's no need to run CI checks on an OWNERS file change.

/override ci/prow/e2e-aws-olm

PR #273, opened 2022-03-30
Ensure that the $ARTIFACTS_DIR variable is respected

timflannagan:
Ensure the `make e2e/olm ARTIFACTS_DIR=...` works when running the upstream OLM e2e suite in downstream CI environments.


timflannagan:
Looks like we may need some downstream-only changes to handle injecting the $ARTIFACTS_DIR variable.

timflannagan:
Holding after the latest comment.

/hold

PR #286, opened 2022-04-11
Override enum

awgreene:


awgreene:
Created this PR to see how we could diverge from the github.com/operator-framework/api repository. Although possible, we probably should not.

PR #292, opened 2022-04-25
Sync OLM commits 04/25

timflannagan:
- Update the staging/operator-lifecycle-manager's directory with the commits that have landed upstream.
- Explicitly override the $KUBECTL binary with oc when running the OLM e2e suite checks


dinhxuanvu (commented):
/lgtm

PR #295, opened 2022-04-29
manifests/*: comply to restricted pod security level

s-urbaniak:
Starting from OpenShift 4.11 pod security admission is being activated. In OpenShift the default pod security admission level is going to be `restricted`. This PR fixes workloads from this repository. Concretely, the following violations have been detected:

```
{
  "objectRef": "openshift-operator-lifecycle-manager/deployments/olm-operator",
  "pod-security.kubernetes.io/audit-violations": "would violate PodSecurity \"restricted:latest\": allowPrivilegeEscalation != false (container \"olm-operator\" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities
(container \"olm-operator\" must set securityContext.capabilities.drop=[\"ALL\"]), runAsNonRoot != true (pod or container \"olm-operator\" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container \"olm-operator\" must set se
curityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")"
}

{
  "objectRef": "openshift-operator-lifecycle-manager/deployments/package-server-manager",
  "pod-security.kubernetes.io/audit-violations": "would violate PodSecurity \"restricted:latest\": allowPrivilegeEscalation != false (container \"package-server-manager\" must set securityContext.allowPrivilegeEscalation=false), unrestricted cap
abilities (container \"package-server-manager\" must set securityContext.capabilities.drop=[\"ALL\"]), runAsNonRoot != true (pod or container \"package-server-manager\" must set securityContext.runAsNonRoot=true), seccompProfile (pod or containe
r \"package-server-manager\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")"
}

{
  "objectRef": "openshift-operator-lifecycle-manager/deployments/catalog-operator",
  "pod-security.kubernetes.io/audit-violations": "would violate PodSecurity \"restricted:latest\": allowPrivilegeEscalation != false (container \"catalog-operator\" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilit
ies (container \"catalog-operator\" must set securityContext.capabilities.drop=[\"ALL\"]), runAsNonRoot != true (pod or container \"catalog-operator\" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container \"catalog-operat
or\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")"
}

{
[...]

PR #299, opened 2022-05-02
manifests/*: comply to restricted pod security level

timflannagan:
Follow-up to #295 which has additional scripts/* changes.


timflannagan:
Waiting for fixes to the e2e/upgrade checks.

/hold

PR #305, opened 2022-05-13
Sync 05 13

perdasilva:
Upstream Sync. Brings down two bug fixes and a small change to the installation script.
This PR should only merge once QA has verified the bug fixes


perdasilva:
Since it's FF I've split this PR into 2 bug PRs. The rest can be sync'ed later

PR #309, opened 2022-05-24
Bug 2088541: Update catalog source pod security context

perdasilva:
Updates the pod and unpack job security context to be more restrictive


dinhxuanvu (commented):
/lgtm

PR #311, opened 2022-05-30
Registry sync 05 30

perdasilva:
Sync'ing upstream operator-registry changes down


timflannagan:
We're past code freeze, which means no more sync PRs unfortunately.

/close

PR #313, opened 2022-06-02
Update makefile to default to gsed if present

awgreene:
Problem: By default, the Mac OS uses POSIX sed which causes the script
used to generate OLM's manifests to fail with a relatively unhelpful
failure message.

Solution: Update the script used to generate OLM's manifests to run on
systems using POSIX or gnu sed.


awgreene:
@exdx

timflannagan (commented):
I think there's also another distinct alternative:

- Add a `SED` variable to avoid hardcoding the sed binary
- That SED variable defaults to `gsed` (IIRC that's the macos equivalent?) if that executable is present, else use `sed`
- Update any `sed` references with `SED`

Thoughts?

awgreene:
> I think there's also another distinct alternative:
> 
>     * Add a `SED` variable to avoid hardcoding the sed binary
> 
>     * That SED variable defaults to `gsed` (IIRC that's the macos equivalent?) if that executable is present, else use `sed`
> 
>     * Update any `sed` references with `SED`
> 
> 
> Thoughts?

@timflannagan I like this approach, I've updated the PR as you suggested and I saved the current implementation in a [backup branch](https://github.com/awgreene/operator-framework-olm/tree/remove-gnu-sed-req-from-makefile-backup) in case we continue to hit the issue.

timflannagan (approved):

PR #320, opened 2022-06-21
Bug 2076323:  Exclude global catalogs from resolution

exdx:
The resolver is currently configured to always consider resolving from global
catalogs. In some cases, this may not be desirable, and the user would like to
explicitly ignore global catalogs for the purposes of resolution. This change
enables per-namespace exclusion from global catalog sources via an annotation on
existing registry source provider.

A queueinformer for OperatorGroups is added since they are now an input to
resolution via the global catalog exclusion annotation. Namespace resolution
will be triggered on changes to an OperatorGroup, in case the value provided on
that annotation by the user changes.

Updated the error message returned by the cache in case a source has an error
to be more clear.

This PR also pulls in e2e test utilities that were added to write the e2e tests 
in a cleaner fashion, 

Signed-off-by: Daniel Sover <dsover@redhat.com>
Co-authored-by: perdasilva <perdasilva@redhat.com>

Upstream-commit: a538831a6b1dfc023989115df8a789bde15b3e53

Upstream-commit: 97b64e1677721faeb3d0e5cd4add2d3298733309

Upstream-commit: dba08eb24af876ffe6672a2fae25362b9b6b3c74


timflannagan (commented):
I _think_ you need to re-generate the root vendor directory here.

jianzhangbjz:
Verified it and comments left: https://bugzilla.redhat.com/show_bug.cgi?id=2076323#c16 
/label qe-approved

tylerslaton:
🤞🏻 
/lgtm

PR #326, opened 2022-06-23
Support Global Operators in Console

awgreene:
**Problem:** Users rely on Copied CSVs in order to understand which
operators are available in a given namespace. When installing All
Namespace operators, a Copied CSV is created in every namespace which
can place a huge performance strain on clusters with many namespaces.
OLM introduced the ability to disable Copied CSVs for All Namespace mode
operators  in an effort to resolve the performance issues on large clusters,
unfortunately removing the ability for console to communicate which global
operators are available in a given namespace.

**Solution:** The protectedCopiedCSVNamespaces runtime flag can be used to
prevent Copied CSVs from being deleted in certain namespaces even when
Copied CSVs are disabled. OLM has been configured to ensure that Copied CSVs
are protected in the openshift namespace. OLM's manifest yaml has also been
updated to provide authenticated users with the RBAC to view CSVs in the
openshift namespace, which console can then use to communicate which operators
are available globally.

### DIFF
--- a/manifests/0000_50_olm_15-csv-viewer.rbac.yaml
+++ b/manifests/0000_50_olm_15-csv-viewer.rbac.yaml
@@ -1,0 +1,36 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  annotations:
+    rbac.authorization.kubernetes.io/autoupdate: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+  name: copied-csv-viewer
+  namespace: openshift
+rules:
+  - apiGroups:
+      - "operators.coreos.com"
+    resources:
+      - "clusterserviceversions"
+    verbs:
+      - get
+      - list
+      - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  annotations:
+    rbac.authorization.kubernetes.io/autoupdate: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+  name: copied-csv-viewers
+  namespace: openshift
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: copied-csv-viewer
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: Group
+    name: system:authenticated
--- a/scripts/olm-deployment.patch.yaml
+++ b/scripts/olm-deployment.patch.yaml
@@ -10,6 +10,14 @@
     name: RELEASE_VERSION
     value: "0.0.1-snapshot"
 - command: update
+  path: spec.template.spec.containers[0].args[+]
+  value:
+    --protectedCopiedCSVNamespaces
+- command: update
+  path: spec.template.spec.containers[0].args[+]
+  value:
+    openshift
+- command: update
   path: spec.template.spec.containers[*].securityContext
   value:
     allowPrivilegeEscalation: false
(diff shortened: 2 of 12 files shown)


timflannagan (commented):
I'm looking at the second commit changes, and I just want to make sure I'm understanding it correctly:

```yaml
            - --protectedCopiedCSVNamespaces
            - openshift
```

Is "openshift" the right value here, or do we need to specify something like "openshift-operators", "openshift-console", etc.?

timflannagan:
Is this related?

```
{  fail [github.com/openshift/origin/test/extended/authorization/rbac/groups_default_rules.go:232]: Jul 27 20:24:39.743: system:authenticated has extra permissions in namespace "openshift":
{APIGroups:["operators.coreos.com"], Resources:["clusterserviceversions"], Verbs:["get" "list" "watch"]}
Ginkgo exit error 1: exit with code 1}
```

manifests/0000_50_olm_15-csv-viewer.rbac.yaml
```diff
@@ -0,0 +1,36 @@
...
+  name: copied-csv-viewers
+  namespace: openshift
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: copied-csv-viewer
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: Group
+    name: system:authenticated
```
timflannagan, line 36: It looks like this might be related to the CI failures:

```log
{  fail [github.com/openshift/origin/test/extended/authorization/rbac/groups_default_rules.go:232]: Jul 27 20:24:39.743: system:authenticated has extra permissions in namespace "openshift":
{APIGroups:["operators.coreos.com"], Resources:["clusterserviceversions"], Verbs:["get" "list" "watch"]}
Ginkgo exit error 1: exit with code 1}
```

awgreene: I've opened https://github.com/openshift/origin/pull/27326 to address this.

scripts/olm-deployment.patch.yaml
```diff
@@ -9,6 +9,14 @@
...
     name: RELEASE_VERSION
     value: "0.0.1-snapshot"
+- command: update
+  path: spec.template.spec.containers[0].args[+]
+  value:
+    --protectedCopiedCSVNamespaces
+- command: update
+  path: spec.template.spec.containers[0].args[+]
+  value:
+    openshift
```
awgreene, line 19: I kept hitting formatting errors when attempting to add 2 lines to the array with a single update command, if anyone knows how to do this in a single update please chime in.

manifests/0000_50_olm_15-csv-viewer.rbac.yaml
timflannagan: Note: The PR that Alex had linked is now merged. Looking at the e2e-gcp check, it looks like those changes worked, and we're back in business.

staging/operator-lifecycle-manager/pkg/controller/operators/olm/operator.go
```diff
@@ -1299,20 +1301,31 @@ func (a *Operator) syncOLMConfig(obj interface{}) (syncError error) {
...
+			}
+			copiedCSVNamespaces[copiedCSV.GetName()][copiedCSV.GetNamespace()] = struct{}{}
 		}
 
 		csvs, err := a.lister.OperatorsV1alpha1().ClusterServiceVersionLister().ClusterServiceVersions(og.GetNamespace()).List(labels.NewSelector().Add(*nonCopiedCSVRequirement))
 		if err != nil {
 			return err
 		}
 
+		namespaces, err := a.lister.CoreV1().NamespaceLister().List(labels.Everything())
```
exdx, line 1319: This List() might be quite expensive. Is there an implementation where we don't rely on knowing all namespaces on the cluster, just the ones that are in the flag?

exdx: Oh wait, it's hitting the cache. This is likely fine then.

PR #335, opened 2022-07-07
Sync 0707

perdasilva:


operator-lifecycle-manager.blacklist
```diff
@@ -0,0 +1,4 @@
+604825027920cdc52a489669b22f51910cabfa69
+fd3322a2bafedce1821ed45b8a9bcce429126af7
+ef91a92858f22a2510e243efc2ae114272b6ad7b
+c7b6534da9ec82b0266fdd69da8ade80d1a8ff80
```
timflannagan, line 4: Any idea on why we needed to skip these commits?

timflannagan: works for me. it might be nice to track that context down $somewhere for our future selves?

PR #345, opened 2022-07-29
Explicitly override the global catalog namespace 

timflannagan:
Update the staging OLM testing suite, and override the E2E_CATALOG_NS and configure the global catalog namespace to the openshift-marketplace namespace. This should fix the permafailing global catalog exclusion tests as they were previously pointing to the wrong downstream namespace.

### DIFF
--- a/staging/operator-lifecycle-manager/test/e2e/catalog_exclusion_test.go
+++ b/staging/operator-lifecycle-manager/test/e2e/catalog_exclusion_test.go
@@ -49,7 +49,7 @@ var _ = Describe("Global Catalog Exclusion", func() {
 		globalCatalog := &v1alpha1.CatalogSource{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      genName("bad-global-catalog-"),
-				Namespace: operatorNamespace,
+				Namespace: globalCatalogNamespace,
 			},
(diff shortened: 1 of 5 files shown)


staging/operator-lifecycle-manager/test/e2e/catalog_exclusion_test.go
```diff
@@ -105,7 +105,7 @@ var _ = Describe("Global Catalog Exclusion", func() {
 		When("the operator group is annotated with olm.operatorframework.io/exclude-global-namespace-resolution=true", func() {
 			// Note: this test flakes on openshift CI but passes running on a local openshift cluster
 			// For some reason, the subscription does not transition to a failed state in time
-			It("[FLAKE] the broken subscription should resolve and have state AtLatest", func() {
+			It("the broken subscription should resolve and have state AtLatest", func() {
```
timflannagan, line 108: Note: this was massaged into the original commit that downstreaming the global catalog exclusion work from upstream. This was added as a downstream-only change that I missed when originally opening the PR.

timflannagan:
Overriding upgrade check. These changes only affect our testing suite, and nothing user facing.

/override ci/prow/e2e-upgrade

PR #366, opened 2022-08-18
[WIP] Label syncer labelling

perdasilva:


staging/operator-lifecycle-manager/pkg/controller/operators/olm/downstream_csv_namespace_labeler_plugin.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/client-go/util/workqueue"
+)
+
+const labelSyncerLabelKey = "security.openshift.io/scc.podSecurityLabelSync"
+
+// systemNSSyncExemptions is the list of namespaces deployed by an OpenShift install
+// payload, as retrieved by listing the namespaces after a successful installation
+// IMPORTANT: The Namespace openshift-operators must be an exception to this rule
+// since it is used by OCP/OLM users to install their Operator bundle solutions.
+var systemNSSyncExemptions = sets.NewString(
```
joelanford, line 26: I take it this is the exact same list as the label syncer's? Any chance we could get them to put that somewhere we could pull in? If not, how do we keep this up-to-date?

staging/operator-lifecycle-manager/pkg/controller/operators/olm/downstream_csv_namespace_labeler_plugin.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+		p.namespaceLister = namespaceInformer.Lister()
+
+		// create a new csv informer and prune status to reduce memory footprint
+		csvNamespaceLabelerInformer := cache.NewSharedIndexInformer(
+			pruning.NewListerWatcher(
+				op.client,
+				namespace,
+				func(opts *metav1.ListOptions) {
+					// opts.LabelSelector = fmt.Sprintf("!%s", v1alpha1.CopiedLabelKey)
```
joelanford, line 130: Why is this commented out?

perdasilva: It was a premature optimization. While it can help reduce the memory footprint, it means a 'not found' error when you try to get it (you might still get csv events from csvs in namespaces that are culled). This felt a bit ambiguous and a potential source of confusion for the next developer. wdyt?

staging/operator-lifecycle-manager/pkg/controller/operators/olm/downstream_csv_namespace_labeler_plugin.go
perdasilva: It is - I just took it from their code for now. But yet, we definitely need to collapse into a single source of truth.

staging/operator-lifecycle-manager/pkg/controller/operators/olm/downstream_csv_namespace_labeler_plugin.go
perdasilva: sorry, I confused myself here. The reason I commented this one out was because I was getting some strange behavior. I wasn't getting events for a non-openshift-* namespace...? I will try again to make sure though.

PR #367, opened 2022-08-18
(psa) restrict olm namespace + remove labels from openshift-operators ns

anik120:
This PR:

1. Adds the enforce:restricted Pod Security Admission labels to the
openshift-operator-lifecycle-manager namespace

2. Adds the enforce:privileged  PSA labels to the openshift-operator 
namespace, that will be removed in a future commit, when another entity 
is present to modify the namespace to set the security of the namespace 
according to the workloads present in the namespace.

### DIFF
--- a/manifests/0000_50_olm_00-namespace.yaml
+++ b/manifests/0000_50_olm_00-namespace.yaml
@@ -3,6 +3,8 @@ kind: Namespace
 metadata:
   name: openshift-operator-lifecycle-manager
   labels:
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: "v1.24"
     openshift.io/scc: "anyuid"
     openshift.io/cluster-monitoring: "true"
   annotations:
@@ -16,7 +18,7 @@ kind: Namespace
 metadata:
   name: openshift-operators
   labels:
-    pod-security.kubernetes.io/enforce: baseline
+    pod-security.kubernetes.io/enforce: privileged
     pod-security.kubernetes.io/enforce-version: "v1.24"
     openshift.io/scc: "anyuid"
   annotations:
--- a/scripts/generate_crds_manifests.sh
+++ b/scripts/generate_crds_manifests.sh
@@ -402,4 +402,5 @@ add_ibm_managed_cloud_annotations "${ROOT_DIR}/manifests"
 find "${ROOT_DIR}/manifests" -type f -exec $SED -i "/^#/d" {} \;
 find "${ROOT_DIR}/manifests" -type f -exec $SED -i "1{/---/d}" {} \;
 
-${YQ} delete --inplace -d'0' manifests/0000_50_olm_00-namespace.yaml 'metadata.labels."pod-security.kubernetes.io/enforce*"'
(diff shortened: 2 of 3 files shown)


manifests/0000_50_olm_00-namespace.yaml
```diff
@@ -3,6 +3,8 @@ kind: Namespace
 metadata:
   name: openshift-operator-lifecycle-manager
   labels:
+    pod-security.kubernetes.io/enforce: restricted
+    pod-security.kubernetes.io/enforce-version: "v1.24"
     openshift.io/scc: "anyuid"
```
camilamacedo86, line 8: /hold 

We just label it to enforce baseline because the pod requires more permissions.
We do not need to enforce as restricted because it is done by default

bparees: all pods running in the openshift-operator-lifecycle-manager are running under the restricted-v2 SCC, thus the PSA enforcement=restricted should be fine.

```
$ oc project
Using project "openshift-operator-lifecycle-manager" on server "https://api.bparees.devcluster.openshift.com:6443".

$ oc get pods -o yaml | grep scc
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
      openshift.io/scc: restricted-v2
```

manifests/0000_50_olm_00-namespace.yaml
```diff
@@ -16,8 +18,6 @@ kind: Namespace
 metadata:
   name: openshift-operators
   labels:
-    pod-security.kubernetes.io/enforce: baseline
-    pod-security.kubernetes.io/enforce-version: "v1.24"
```
bparees, line 20: until we have the OLM logic in place to label "openshift-* namespaces that have operators installed" for labelsyncing, we should probably leave this explicit setting in place (or even set it to privileged), so that operators that are installed in this NS don't get rejected by PSA.

cc @perdasilva

manifests/0000_50_olm_00-namespace.yaml
camilamacedo86: Cool so we merge the changes 🚀

manifests/0000_50_olm_00-namespace.yaml
camilamacedo86: I think needs to be previlged we do not know what will run in openshift-operators. It is the default namespace to install the projects.

anik120: That is a good point. I've modified the PR to go with option 1 and explicitly label it as `privileged`, so that when https://issues.redhat.com/browse/OLM-2695 is ready we have to make a conscious effort to remove any hard-coding of labels in this namespace.

manifests/0000_50_olm_00-namespace.yaml
anik120: /hold cancel

camilamacedo86 (approved):
/lgtm

PR #378, opened 2022-09-06
OCPBUGS-858: Package Server Manager should enforce expected csv values

awgreene:
**Problem:**
The package-server-manifest (PSM) is a downstream specific component responsible for reconciling the package-server csv, which is defined [here](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/pkg/manifests/csv.yaml) as a yaml file. When the PSM reconciles the package-server csv, it:
- Attempts to generate the base of the expected CSV [here](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/pkg/package-server-manager/controller.go#L76-L84)
- Passes the [reconcile function](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/pkg/package-server-manager/controller.go#L97-L118) as the `mutateFn` parameter and the expectedCSV  as the `obj` parameter into into controller-runtime's [createOrUpdate function](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/vendor/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil/controllerutil.go#L195-L223). Controller runtime's `createOrUpdate function` will apply the `mutateFn` function against:
- The `obj` you passed in if it does not already exist.
OR
- The existing on cluster object.

In the case where the `mutateFn` is applied to the applied to the existing object, changes made to the [package-server csv yaml](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/pkg/manifests/csv.yaml) are ignored because the `mutateFn` (ie the `reconcileCSV` function) doesn't consider changes to the manifest.

**Solution:**
Update the `reconcileCSV` function to:
- Ensure that expected annotations and labels are present on the csv.
- Ensure that the expected spec is present on the csv.

### DIFF
--- a/pkg/package-server-manager/config.go
+++ b/pkg/package-server-manager/config.go
@@ -11,6 +11,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/pointer"
+
+	"github.com/openshift/operator-framework-olm/pkg/manifests"
 )
 
 func getReplicas(ha bool) int32 {
@@ -61,8 +63,61 @@ func getTopologyModeFromInfra(infra *configv1.Infrastructure) bool {
 }
 
 // ensureCSV is responsible for ensuring the state of the @csv ClusterServiceVersion custom
+// resource matches that of the codified defaults and high availability configurations, where
+// codified defaults are defined by the csv returned by the manifests.NewPackageServerCSV
+// function.
+func ensureCSV(log logr.Logger, image string, csv *olmv1alpha1.ClusterServiceVersion, highlyAvailableMode bool) (bool, error) {
+	expectedCSV, err := manifests.NewPackageServerCSV(
+		manifests.WithName(csv.Name),
+		manifests.WithNamespace(csv.Namespace),
+		manifests.WithImage(image),
+	)
+	if err != nil {
+		return false, err
+	}
+
+	ensureCSVHighAvailability(image, expectedCSV, highlyAvailableMode)
+
+	var modified bool
+
+	for k, v := range expectedCSV.GetLabels() {
+		if csv.GetLabels() == nil {
+			csv.SetLabels(make(map[string]string))
+		}
+		if vv, ok := csv.GetLabels()[k]; !ok || vv != v {
+			log.Info("setting expected label", "key", k, "value", v)
+			csv.ObjectMeta.Labels[k] = v
+			modified = true
+		}
+	}
+
+	for k, v := range expectedCSV.GetAnnotations() {
+		if csv.GetAnnotations() == nil {
+			csv.SetAnnotations(make(map[string]string))
+		}
+		if vv, ok := csv.GetAnnotations()[k]; !ok || vv != v {
+			log.Info("setting expected annotation", "key", k, "value", v)
+			csv.ObjectMeta.Annotations[k] = v
+			modified = true
+		}
+	}
+
+	if !reflect.DeepEqual(expectedCSV.Spec, csv.Spec) {
+		log.Info("updating csv spec")
+		csv.Spec = expectedCSV.Spec
+		modified = true
+	}
+
+	if modified {
+		log.V(3).Info("csv has been modified")
+	}
+
+	return modified, err
+}
+
+// ensureCSVHighAvailability is responsible for ensuring the state of the @csv ClusterServiceVersion custom
 // resource matches the expected state based on any high availability expectations being exposed.
-func ensureCSV(log logr.Logger, image string, csv *olmv1alpha1.ClusterServiceVersion, highlyAvailableMode bool) bool {
+func ensureCSVHighAvailability(image string, csv *olmv1alpha1.ClusterServiceVersion, highlyAvailableMode bool) {
 	var modified bool
 
 	deploymentSpecs := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
@@ -70,52 +125,29 @@ func ensureCSV(log logr.Logger, image string, csv *olmv1alpha1.ClusterServiceVer
 
 	currentImage := deployment.Template.Spec.Containers[0].Image
 	if currentImage != image {
-		log.Info("updating the image", "old", currentImage, "new", image)
 		deployment.Template.Spec.Containers[0].Image = image
 		modified = true
 	}
 
 	expectedReplicas := getReplicas(highlyAvailableMode)
 	if *deployment.Replicas != expectedReplicas {
-		log.Info("updating the replica count", "old", deployment.Replicas, "new", expectedReplicas)
 		deployment.Replicas = pointer.Int32Ptr(expectedReplicas)
 		modified = true
 	}
 
 	expectedRolloutConfiguration := getRolloutStrategy(highlyAvailableMode)
 	if !reflect.DeepEqual(deployment.Strategy.RollingUpdate, expectedRolloutConfiguration) {
-		log.Info("updating the rollout strategy")
 		deployment.Strategy.RollingUpdate = expectedRolloutConfiguration
 		modified = true
 	}
 
 	expectedAffinityConfiguration := getAntiAffinityConfig(highlyAvailableMode)
 	if !reflect.DeepEqual(deployment.Template.Spec.Affinity, expectedAffinityConfiguration) {
-		log.Info("updating the pod anti-affinity configuration")
 		deployment.Template.Spec.Affinity = expectedAffinityConfiguration
 		modified = true
 	}
 
 	if modified {
-		log.V(3).Info("csv has been modified")
 		csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec = *deployment
 	}
-
-	return modified
-}
-
-func validateCSV(log logr.Logger, csv *olmv1alpha1.ClusterServiceVersion) bool {
-	deploymentSpecs := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
-	if len(deploymentSpecs) != 1 {
-		log.Info("csv contains more than one or zero nested deployment specs")
-		return false
-	}
-
-	deployment := &deploymentSpecs[0].Spec
-	if len(deployment.Template.Spec.Containers) != 1 {
-		log.Info("csv contains more than one container")
-		return false
-	}
-
-	return true
 }
--- a/pkg/package-server-manager/controller.go
+++ b/pkg/package-server-manager/controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"sync"
 
 	"github.com/go-logr/logr"
@@ -85,11 +86,12 @@ func (r *PackageServerCSVReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	res, err := controllerutil.CreateOrUpdate(ctx, r.Client, required, func() error {
 		return reconcileCSV(r.Log, r.Image, required, highAvailabilityMode)
 	})
+
+	log.Info("reconciliation result", "res", res)
 	if err != nil {
 		log.Error(err, "failed to create or update the packageserver csv")
 		return ctrl.Result{}, nil
 	}
-	log.Info("reconciliation result", "res", res)
 
 	return ctrl.Result{}, nil
 }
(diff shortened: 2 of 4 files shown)


awgreene:
WIP
/hold

pkg/package-server-manager/controller.go
```diff
@@ -98,22 +99,31 @@ func reconcileCSV(log logr.Logger, image string, csv *olmv1alpha1.ClusterService
...
+	if err != nil {
+		return err
 	}
-	if !ensureCSV(log, image, csv, highAvailabilityMode) {
+
+	if !ensureCSV(log, image, expectedCSV, highAvailabilityMode) {
 		log.V(3).Info("no further updates are necessary to the packageserver csv")
 	}
 
+	// Ensure defaults
```
tylerslaton, line 116: nit: Maybe expand a bit more on what it means to ensure these defaults here

tylerslaton (commented):
/lgtm

tylerslaton (approved):

anik120 (commented):
Trying to understand this: 

So what this PR is saying is: any CSV >= v0.18.3 was being flagged as an invalid CSV, because one of the hardcoded conditions inside `validateCSV` was no longer correct (i.e only one deployment spec, and one container for the deployment). So the fix is to remove that check, which was setting values back to default, i.e the CSV back to v0.17.0?

awgreene:
> Trying to understand this:
> 
> So what this PR is saying is: any CSV >= v0.18.3 was being flagged as an invalid CSV, because one of the hardcoded conditions inside validateCSV was no longer correct (i.e only one deployment spec, and one container for the deployment). So the fix is to remove that check, which was setting values back to default, i.e the CSV back to v0.17.0?

Great question Anik, I'll be sure to update the commit message and PR description to include this information, but:

The package-server-manifest (PSM) is a downstream specific component responsible for reconciling the package-server csv, which is defined [here](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/pkg/manifests/csv.yaml) as a yaml file. When the PSM reconciles the package-server csv, it:
- Attempts to generate the base of the expected CSV [here](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/pkg/package-server-manager/controller.go#L76-L84)
- Passes the [reconcile function](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/pkg/package-server-manager/controller.go#L97-L118) as the `mutateFn` parameter and the expectedCSV  as the `obj` parameter into into controller-runtime's [createOrUpdate function](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/vendor/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil/controllerutil.go#L195-L223). Controller runtime's `createOrUpdate function` will apply the `mutateFn` function against:
1. the `obj` you passed in if it does not already exist
OR
2. the existing on cluster object

In the case where the `mutateFn` is applied to the applied to the existing object, changes made to the [package-server csv yaml](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/pkg/manifests/csv.yaml) are ignored because the mutateFn (ie the reconcileCSV function) doesn't apply them.

awgreene:
Updated the PR description and the commit message.

pkg/package-server-manager/controller.go
awgreene: Done.

pkg/package-server-manager/controller.go
```diff
@@ -98,22 +99,35 @@ func reconcileCSV(log logr.Logger, image string, csv *olmv1alpha1.ClusterService
...
 
+	for k, v := range expectedCSV.GetLabels() {
+		csv.ObjectMeta.Labels[k] = v
+	}
+
+	for k, v := range expectedCSV.GetAnnotations() {
+		csv.ObjectMeta.Annotations[k] = v
+	}
+
+	csv.Spec = expectedCSV.Spec
```
grokspawn, line 129: So basically, unless our narrowly-defined [validateCsv](https://github.com/openshift/operator-framework-olm/blob/fceaf8aba28fb5c9c8f636c49b0fe3119d7d348d/pkg/package-server-manager/config.go#L107-L121) failed, we would not consider what the default contained.  

And then you're adding stickiness to the existing labels and annotations, which should address the original complaint.  

The only question I have is that in the old code we accumulated to `csv.Spec` before invoking `ensureCSV` and now we do it afterwards.  Was the reordering intentional?

awgreene: That seems reasonable to me

awgreene: I think we'd want to add the annotations and labels to ensureCSV as well.

awgreene: To be clear, I generated the expectedCSV first so the ensureCSV function could set the expected values, but it seems reasonable to move the creation of the expectedCSV within the ensureCSV method.

awgreene: And the expectedCSV is passed into the ensureCSV method before the cab.Spec is set

grokspawn, line 129: Grr.... I wanted to post my understanding as a suggestion, so you could easily tell me to punt if I'm off-base here.  
Where I'm not clear is if all the action is in the `ensureCSV` method, and action includes scope over all the accumulations we're making.  
It /seems/ so, in which case the last thing we want to do is invoke `ensureCSV`. 

Here's my best attempt at listing what I think should happen here:

```

	// The Package Server CSV is defined within this repo as a yaml manifest which
	// acts as a "base" of the CSV we expect. We then apply the highAvailability
	// configurations with the `ensureCSV` method to the expectedCSV object. The
	// csv passed into this function then has the expected labels, annotations,
	// and spec set.
	expectedCSV, err := manifests.NewPackageServerCSV(
		manifests.WithName(csv.Name),
		manifests.WithNamespace(csv.Namespace),
		manifests.WithImage(image),
	)
	if err != nil {
		return err
	}

	for k, v := range expectedCSV.GetLabels() {
		csv.ObjectMeta.Labels[k] = v
	}

	for k, v := range expectedCSV.GetAnnotations() {
		csv.ObjectMeta.Annotations[k] = v
	}

	csv.Spec = expectedCSV.Spec
	
	if !ensureCSV(log, image, csv, highAvailabilityMode) {
		log.V(3).Info("no further updates are necessary to the packageserver csv")
	}
	```

Comment (awgreene) [reply pkg/package-server-manager/controller.go:129]: Hey! I've added the tests you requested!

Comment (awgreene) [reply pkg/package-server-manager/controller.go:129]: Hey there @grokspawn, thanks for the review! I've moved all the logic within the ensure method so that it:
- Generates the csv from the manifest defined in code
- Applies HA configurations to the manifest
- Ensures that the proper annotations/labels exists
- That the actual spec matches the expected spec

**Comment by awgreene** [comment]
I've added tests to ensure that changes to the CSV's spec are reverted.
/unhold

File: pkg/package-server-manager/config.go [hunk 3]
```diff
@@ -61,61 +63,85 @@ func getTopologyModeFromInfra(infra *configv1.Infrastructure) bool {
...
+		log.V(3).Info("csv has been modified")
+	}
+
+	return modified, err
+}
+
+// ensureCSVHighAvailability is responsible for ensuring the state of the @csv ClusterServiceVersion custom
 // resource matches the expected state based on any high availability expectations being exposed.
-func ensureCSV(log logr.Logger, image string, csv *olmv1alpha1.ClusterServiceVersion, highlyAvailableMode bool) bool {
+func ensureCSVHighAvailability(image string, csv *olmv1alpha1.ClusterServiceVersion, highlyAvailableMode bool) {
```
awgreene, line 114: Note: Now that we ALWAYS apply this against the CSV in the manifest, it will ALWAYS result in a modified object, so I decided to remove the returned bool.

anik120:
I wonder if there's a way to launch a 4.9 cluster with this PR. It'd be ideal to do an upgrade testing from 4.9->4.12 with this PR, and make sure that for each OCP version, the correct CSV version is being deployed on upgrade.

pkg/package-server-manager/controller_test.go
```diff
@@ -133,83 +163,104 @@ func TestEnsureCSV(t *testing.T) {
...
+		expectedErr  error
+	}
+
 	tt := []struct {
 		name            string
 		inputCSV        *olmv1alpha1.ClusterServiceVersion
 		expectedCSV     *olmv1alpha1.ClusterServiceVersion
 		highlyAvailable bool
-		want            bool
+		want            wanted
```
perdasilva, line 176: all test cases seem to set expectedErr to nil. Is it worth exercising the failure path as well?
Seems like it only happens when there are issues deserializing the csv

pkg/package-server-manager/controller.go
```diff
@@ -98,19 +100,13 @@ func reconcileCSV(log logr.Logger, image string, csv *olmv1alpha1.ClusterService
...
-		}
-		csv.Spec = tmp.Spec
+
+	modified, err := ensureCSV(log, image, csv, highAvailabilityMode)
+	if err != nil {
+		return fmt.Errorf("error ensuring CSV: %v", err)
 	}
-	if !ensureCSV(log, image, csv, highAvailabilityMode) {
+
+	if modified {
```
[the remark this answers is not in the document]
awgreene: Yes, updated.

pkg/package-server-manager/controller_test.go
awgreene: That's correct, all tests expect no error from this method. The request to add an error seems reasonable, but the only state where ensureCSV can return an error is when it has [an error unmarshalling the packageserver.yaml](https://github.com/openshift/operator-framework-olm/blob/2149aebcc71367e6fba8f5416374917dae1e6a1c/pkg/manifests/manifests.go#L18-L25), which is hardcoded and not exposed via the function. While we could mock the function or update it to take a configurable path, it seemed like overkill and I didn't want to increase the scope of this PR given that it will need to be backported to 4.9.

If you feel strongly that we should test this path, I can update it when I'm back from vacation next week.

anik120: my preference would be to not increase the scope of this PR either, specially with the need to backport in mind. We could always have a follow up PR for the test

pkg/package-server-manager/controller_test.go
```diff
@@ -72,10 +74,39 @@ func intOrStr(val int) *intstr.IntOrString {
...
+		csv.Spec.Description = description
+	}
+}
+
+func withAffinity(affinity *corev1.Affinity) func(*olmv1alpha1.ClusterServiceVersion) {
+	return func(csv *olmv1alpha1.ClusterServiceVersion) {
+		csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec.Template.Spec.Affinity = affinity
+	}
+}
+func withRolloutStrategy(strategy *appsv1.RollingUpdateDeployment) func(*olmv1alpha1.ClusterServiceVersion) {
```
awgreene, line 96: Shoot, @grokspawn could you change this to the name of the field?

pkg/package-server-manager/config.go
```diff
@@ -61,61 +63,85 @@ func getTopologyModeFromInfra(infra *configv1.Infrastructure) bool {
...
+	)
+	if err != nil {
+		return false, err
+	}
+
+	ensureCSVHighAvailability(image, expectedCSV, highlyAvailableMode)
+
+	var modified bool
+
+	for k, v := range expectedCSV.GetLabels() {
```
awgreene, line 83: We should add a check that the csv's annotations and label maps are not nil @grokspawn. If the are nil, set them to new maps and populate with the following for loops.

awgreene: Something like

'''
if csv.GetAnnotations() == nil {
   csv.SetAnnotations(make(map[string]string))
}
'''

awgreene (commented):
Last minute additions

pkg/package-server-manager/controller_test.go
awgreene: Maybe withRollingUpdateStrategy? Open to suggestions

grokspawn: How about `withRollingUpdateStrategy`?

grokspawn:
@ qe, please take a look at the reproduction steps in [OCPBUGS-867](https://issues.redhat.com/browse/OCPBUGS-867) for a (relatively) easy verification.

pkg/package-server-manager/controller_test.go
grokspawn: Funny that your comment came in after mine, but at least great minds think alike. :)

jianzhangbjz:
> I wonder if there's a way to launch a 4.9 cluster with this PR. It'd be ideal to do an upgrade testing from 4.9->4.12 with this PR, and make sure that for each OCP version, the correct CSV version is being deployed on upgrade.

Yes, but I guess no. In fact, I build a payload with this PR by using cluster-bot, and upgrade OCP 4.10.32 to it forcefully, it works well.

dinhxuanvu (commented):
/lgtm

grokspawn:
/hold cancel
got the qe-approved label 😄

jianzhangbjz:
LGTM, details see: https://issues.redhat.com/browse/OCPBUGS-867 and https://issues.redhat.com/browse/OCPBUGS-858

PR #387, opened 2022-09-20
[release-4.10] OCPBUGS-1525: Package Server Manager should enforce expected csv values

anik120:
**Problem:**
The package-server-manifest (PSM) is a downstream specific component responsible for reconciling the package-server csv, which is defined in code as a yaml file. When the PSM reconciles the package-server csv, it:
- Attempts to generate the base of the expected CSV from the yaml mentioned above.
- Passes the reconcileCSV function as the mutateFn parameter and the expectedCSV  as the `obj` parameter into into controller-runtime's createOrUpdate function.

Controller runtime's createOrUpdate function will apply the mutateFn function against:
- The `obj` you passed in if it does not already exist. OR
- The existing on cluster object.

In the case where the mutateFn is applied to the applied to the existing object, changes made to the package-server csv yaml are ignored because the mutateFn (ie the reconcileCSV function) doesn't consider changes to the manifest.

**Solution:**
Update the reconcileCSV function to:
- Ensure that expected annotations and labels are present on the csv.
- Ensure that the expected spec is present on the csv.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/openshift/operator-framework-olm
 go 1.17
 
 require (
+	github.com/blang/semver/v4 v4.0.0
 	github.com/go-bindata/go-bindata/v3 v3.1.3
 	github.com/go-logr/logr v0.4.0
 	github.com/golang/mock v1.6.0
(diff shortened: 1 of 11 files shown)


timflannagan (approved):
/uncc

anik120:
looks like the e2e-aws-olm job was removed one week before this PR was opened https://github.com/openshift/release/commit/c5e100bbf0d5eb93f3366a3891c2a4edd9c60860

which is why ci isn't responding to our test commands. Overriding this test since it's already removed. 

/override e2e-aws-olm

anik120:
/override e2e-aws-olm

ref: https://github.com/openshift/operator-framework-olm/pull/387#issuecomment-1384311329

PR #391, opened 2022-09-22
OCPBUGS-1912: Move opm diff to oc mirror

grokspawn:


timflannagan (approved):

timflannagan (commented):
It looks like we're introducing a staging/*/vendor file in one of the cherry picked commits:

```bash
$ git show HEAD~1 --stat staging/operator-registry/ | grep vendor
 staging/operator-registry/vendor/modules.txt       | 1313 ++++++++
```

Which is causing the registry unit test failures. Manually removing this nested vendor directory produces a local diff, and re-running the `make unit/registry` resulted in a successful run.

/lgtm cancel

timflannagan (approved):

PR #394, opened 2022-09-29
OCPBUGS-1088: Run collect-profile pod on management nodes

awgreene:
Problem:
The collect-profiles pod spec is missing the
"target.workload.openshift.io/management:" annotation. As a result when the workload partitioning feature is enabled on SNO, this pod resources will not get mutated and pinned to the reserved cpuset.

Solution:
Add the target.workload.openshift.io/management:" annotation to the collect-profiles pod.

### DIFF
--- a/manifests/0000_50_olm_07-collect-profiles.cronjob.yaml
+++ b/manifests/0000_50_olm_07-collect-profiles.cronjob.yaml
@@ -12,6 +12,9 @@ spec:
   jobTemplate:
     spec:
       template:
+        metadata:
+          annotations:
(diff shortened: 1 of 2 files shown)


timflannagan (approved):

awgreene:
/hold
I don't see the annotation on the collect profiles pod, I need to check if this is intended.

awgreene:
I spoke with Brent Rowsell (who submitted the original bug) and found out that the behavior observed in my previous comment is expected on standard openshift clusters.

/unhold

perdasilva:
@awgreene I modified the jira issue to fix the jira validation here. This means we'll probably have to clone it and backport to 4.12, etc.

PR #395, opened 2022-09-29
OCPBUGS-1684: Set cpu limit on collect-profiles job

awgreene:
Problem:
There have been reports of high cpu utilization by the collect profiles pod on customer cluster.

Solution:
Set a cpu limit to prevent extreme cpu utilization.

### DIFF
--- a/manifests/0000_50_olm_07-collect-profiles.cronjob.yaml
+++ b/manifests/0000_50_olm_07-collect-profiles.cronjob.yaml
@@ -47,6 +47,8 @@ spec:
                 requests:
                   cpu: 10m
                   memory: 80Mi
+                limits:
+                  cpu: 100m
           volumes:
             - name: config-volume
               configMap:
--- a/scripts/generate_crds_manifests.sh
+++ b/scripts/generate_crds_manifests.sh
@@ -321,6 +321,8 @@ spec:
               requests:
                 cpu: 10m
                 memory: 80Mi
+              limits:
+                cpu: 100m
           volumes:
           - name: config-volume
             configMap:


awgreene:
We need to consider the potential impact of this change by checking how this affects the job.
/hold

awgreene:
Looks like the resource limit didn't introduce much  of a performance impact:
```
$ k get pods -o yaml | grep limit -B2 -A1
      name: collect-profiles
      resources:
        limits:
          cpu: 100m
--
      name: collect-profiles
      resources:
        limits:
          cpu: 100m
--
      name: collect-profiles
      resources:
        limits:
          cpu: 100m
          
$ k get jobs
NAME                        COMPLETIONS   DURATION   AGE
collect-profiles-27742680   1/1           11s        35m
collect-profiles-27742695   1/1           9s         20m
collect-profiles-27742710   1/1           13s        5m43s
```

grokspawn:
We had also discussed the possibility of disabling this pod by default.  Do you want to keep that outside this PR?

fgiloux:
My 2 cents:
IMHO this change does not address the root cause. The added value of setting a limit is very low.
With the cpu request set to 10m the collect-profiles job has been able to make high CPU consumption only because there were idle cycles. If there was CPU contention on the node processes with higher requests would have had the cycles (limit does not play a role in allocation of extra cycles: the ones that have not been requested).
Setting a limit will mean that it will just take more time for completion, possibly still using the same amount of CPU cycles. The outcome may only be different if there is a timing issue causing unnecessary loops. This brings back to the first point that this PR does not identify and address the root cause.

PR #396, opened 2022-09-30
OCPBUGS-1428: fix service account token secret reference (#2862)

awgreene:
Problem: The filterSecretsBySAName function attempts to identify all service account token secrets related to a serviceAccount. To do so, the filterSecretsBySAName function uses a range-for loop to iterate over entries in the secrets argument. If a valid service account token secret is found, a pointer to the range-for loop's value variable is added to a map of results. Unfortunately, if a valid entry is found in the middle of the list of secrets, the value returned by the range-for loop is updated, causes the entry in the map to change.

Solution: Add a pointer to the actual secret instead of the range-for loop's value variable.

Upstream-commit: caab6c52ec532dc82c7178eebb0377bd80d1e82a


dinhxuanvu (approved):
/lgtm

PR #399, opened 2022-10-05
Attempt workaround for https://issues.redhat.com/browse/CLOUDBLD-11068

bparees:
this is based on the fact that:

1) cluster-policy-controller is able to dep pod security admission and build ok:
https://github.com/openshift/cluster-policy-controller/blob/6dba7561af11e0a949bbeaa57958b195b98b8d0a/go.mod#L140

2) olm doesn't actually need pod security admission, it needs one file from cluster-policy-controller.  That seems to drag the transitive dep in during the brew build, but go mod vendor does not pull it in.  Which makes me think brew is resolving the dep differently and not resolving to the right thing.  Hoping the explicit replaces helps that.

that said, cluster-policy, since it actually deps psa, also has this:

https://github.com/openshift/cluster-policy-controller/blob/6dba7561af11e0a949bbeaa57958b195b98b8d0a/go.sum#L952-L953

which olm of course doesn't, because it doesn't actually need the dep.


timflannagan:
Nothing quite like non-deterministic unit tests.

/retest

bparees:
it's basically impossible for this to break anything in CI since it doesn't actually change any of the code that's being built

grokspawn (approved):

PR #404, opened 2022-10-24
[WIP] OCPBUGS 2782: Update service deletion to retry with timeout

perdasilva:
Signed-off-by: perdasilva [perdasilva@redhat.com](mailto:perdasilva@redhat.com)

Description of the change:
Adds a retry with timeout when attempting to delete a service during installation/upgrade.
This is a small mitigation for the bug linked in the title. The real fix for this would be to make install plan retryable. But given the efforts towards v1, it's probably not worth it.

Motivation for the change:
In the operator upgrade process, any services will be deleted and re-created. Sometimes these services can have finalizers attached that may delay the deletion of the resource and cause the upgrade to fail.


perdasilva:
closing and this does not seem like the direction we want to take

PR #405, opened 2022-10-27
WIP: fix operator controller

awgreene:


timflannagan (commented):
Talked in slack, but just recording the context from that conversation. Can we use a sort.Slice implementation to avoid having to update the o-f/api repository, which should make it easier to backport these changes?

awgreene:
>Talked in slack, but just recording the context from that conversation. Can we use a sort.Slice implementation to avoid having to update the o-f/api repository, which should make it easier to backport these changes?

Yep! Great suggestion. Updated.

PR #417, opened 2022-11-30
OCPBUGS-3881: Revert CatalogSource security changes for PSA

perdasilva:
Signed-off-by: perdasilva <perdasilva@redhat.com>


awgreene:
Note: The verify test will never change as this revert change is being made downstream only.

anik120:
Carrying over comment from OCPBUGS-3881:

Here's another, small change option to ensure IBM clusters are not impacted, instead of the wholesale revert all PSA changes + re-enable them at a future release :
Introduce logic downstream to add spec.GrpcPodConfig.SecurityContextConfig:legacy for any CatalogSource created in the `openshift-marketplace` namespace
Mark `openshift-marketplace` namespace as PSA `privileged`
That way in 4.13(or whichever release we decide is ready), we just have to revert 1 in downstream (few lines of code) + determine if it's safe to mark `openshift-marketplace` as `restricted` again (it not `required`, but is `recommended` however to run workload in `restricted` mode)

Reverting PSA changes wholesale has cascading effects on the work that's already in motion (eg labelSyncer controller, doc changes etc) that all needs to be audited again

joelanford:
There was a discussion here about the suggested changes we could make in 4.12: https://coreos.slack.com/archives/C04C5P2PMR6/p1669069382681739

This PR doesn't align with that discussion. Is that because we've moved on or arrived at a different consensus? Can we get consensus on the plan?

perdasilva:
closing this PR as it doesn't meet what original plan T_T

PR #420, opened 2022-12-10
OCPBUGS-4757: Default to legacy psa settings (#273)

awgreene:
The catalogSource api was recently updated to support running the associated pod in a "restricted" workflow as defined by the Pod Security Admission controller. By default, the catalogSource pods have been configured to run in "restricted" mode, which is disruptive to customers managing and deploying their own catalogSources as they need to rebuild their catalogs to run in "restricted" mode if a namespace is marked as "restricted".

In an effort to provide users with a bit more time to rebuilt their catalogSources, this change configures catalogSources to run in "legacy" mode by default. A series of other changes will be made to update the namespaces associated with marketplace and olm  to support catalogSources running in "legacy" mode by default.


jianzhangbjz:
Verify failed, details: https://issues.redhat.com/browse/OCPBUGS-4757
/hold

awgreene:
This PR will need the upstream changes introduced here: https://github.com/operator-framework/operator-lifecycle-manager/pull/2906

awgreene:
@jianzhangbjz this should be good for a retest.

staging/operator-lifecycle-manager/deploy/chart/values.yaml
```diff
@@ -2,8 +2,12 @@ rbacApiVersion: rbac.authorization.k8s.io
 namespace: operator-lifecycle-manager
 # see https://kubernetes.io/docs/concepts/security/pod-security-admission/ for more details
 namespace_psa: 
-  enforceLevel: restricted
+  enforceLevel: baseline
   enforceVersion: latest
+  auditLevel: restricted
+  auditVersion: latest
+  warnLevel: restricted
+  warnVersion: latest
```
perdasilva, line 10: are we sure we want latest here? there was a reason we pinned the version @anik120 do you remember it?

awgreene: These are the upstream chart values, we actually unset the warn and audit bits [here](https://github.com/openshift/operator-framework-olm/pull/420/files#diff-e8033c7c4bf9c1999d5b66faebd942551d51544ef15dd0d43f7efca074981684R411-R412) as the namespaces are restricted.

jianzhangbjz:
Retest it and It works well, details: https://issues.redhat.com/browse/OCPBUGS-4757
/unhold

PR #422, opened 2022-12-14
Updates to OWNERS file

grokspawn:
Tried to prune out folks no longer on the team and add folks who I knew should be on  (like me 😄 ).  If I missed any, please let me know.


anik120 (approved):
/approve

PR #429, opened 2023-01-16
OCPBUGS-5523: Catalog, fatal error: concurrent map read and map write

dtfranz:
Problem:
The metrics pkg uses a shared map var to keep a record of Prometheus counters from Subscription objects, which is not protected from concurrent read/write access. This can result in container restarts. 

Solution: 
This PR adds a mutex to lock and unlock before and after each map access instance. Any number of requests to **read** from the map may happen concurrently, but when a **write** operation occurs nothing else can access the map until the **write** is finished.

Motivation for the change:

OCPBUGS-5523

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/metrics/metrics.go
+++ b/staging/operator-lifecycle-manager/pkg/metrics/metrics.go
@@ -1,6 +1,7 @@
 package metrics
 
 import (
+	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


awgreene:
@dtfranz seems like a gcp failure:
```
{  failed to acquire lease for "gcp-quota-slice": resources not found}
```

tmshort (approved):
/lgtm

PR #441, opened 2023-02-06
OCPBUGS-7086: cherry-pick pull request refactor FBC caching (#1051) f…

grokspawn:
…rom openshift/master

refactor FBC caching (#1051)

This commit improves maintainability and extensibility of the FBC caching used to serve the GRPC API.

It:
- introduces a new cache package and interface, which defines primitives for checking cache integrity, building the cache, and loading the cache.
- moves the existing registry.Querier to an implementation of the cache interface, called cache.JSON

This refactor also resolves two outstanding issues:
1. The current code contains a bug that causes panics when certain failures occur loading the cache.
2. The current code is hardcoded to always rebuild the cache if it is unreadable or its digest doesn't contain the expected value.

This commit:
1. Refactors and removes the cases the led to the panic situation
2. Adds a new flag, --cache-enforce-integrity, that causes `opm serve` to exit with a failure if the cache is unreadable or its digest doesn't contain the expected value. This is useful in production contexts when one always expects the cache to be pre-populated correctly.

Upstream-commit: 562cf9c9bc649f1df2bbae26f8b543529499b60f


Xia-Zhao-rh:
LGTM, verified.
Detail: https://issues.redhat.com/browse/OCPBUGS-7086

PR #454, opened 2023-02-20
OCPBUGS-7769: [release-4.12] update cluster policy operator dependency

perdasilva:
Update the CPC dependency to fix an ART issue that is impeding OLM from being released


bparees:
nothing for QE to validate here so adding cherry-pick-approved

PR #459, opened 2023-02-22
add .status.relatedObjects to Operator to be compatible with oc adm inspect

deads2k:
oc adm inspect will automatically traverse these references and collect the instances.  See https://github.com/openshift/oc/blob/49b1e9f681b3dbdd4f43ca3d288ac64f4c681e2d/pkg/cli/admin/inspect/resource.go#L271

@awgreene This is related to the must-gather thread from today.


ankitathomas:
This feels like it should be a change on the [upstream api](https://github.com/operator-framework/api) repository

PR #461, opened 2023-02-27
OPECO-2737: Veneer template rename

grokspawn:
Pulling along several commits in order to make this a simpler cherry-pick.


grokspawn:
/hold 
This is getting increasingly sticky, both in terms of the number of dependent commits that need to be brought in (looks complete now?) and the nested docker/podman actions which currently fail CI.  
I think we need to explore taking up an existing issue to avoid the nesting, which would make this downstream go more smoothly.

grokspawn:
/hold cancel

Now downstreaming the removal of nested docker execution which should enable e2e.

grokspawn:
/override ci/prow/verify
Adding some upstream OWNER updates that we don't plan to ever downstream, and verify hates the 'downstream-only' nature of it. :(

grokspawn:
This is a renaming event across all alpha functionality, seeking parity with upstream.  Applying labels
/label qe-approved
/label docs-approved
/label px-approved

grokspawn:
RE-overriding ci/prow/verify due to downstream-only OWNER changes, since this has to be done in both review and pre-commit phases. :( 

/override ci/prow/verify
Blacklisting some upstream OWNER updates to that we don't plan to ever downstream, and verify hates the 'downstream-only' nature of it. :(

PR #463, opened 2023-03-07
OPECO-2646: exclude bundles with `olm.deprecated` property when rendering

perdasilva:
* exclude bundles with `olm.deprecated` property when rendering
* Exclude bundles based on the property type and not value


rashmigottipati (approved):
/lgtm

everettraven (approved):
/lgtm

PR #468, opened 2023-03-09
OCPBUGS-6016: UpdateStrategy RegistryPoll with nil Interval

dtfranz:
Adds protection against a nil-pointer panic when an UpdateStrategy with nil value for RegistryPoll Interval is supplied. Also adds a unit test to ensure that the code does not panic but instead returns an error when this situation is encountered.

Upstream-commit: 4bae06a5d2230e62328f5d82ce36c09a7522cea8

Bug: OCPBUGS-6016

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/operators/catalog/operator.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/operators/catalog/operator.go
@@ -749,16 +749,19 @@ func (o *Operator) syncRegistryServer(logger *logrus.Entry, in *v1alpha1.Catalog
 	logger.Debug("ensured registry server")
 
 	// requeue the catalog sync based on the polling interval, for accurate syncs of catalogs with polling enabled
-	if out.Spec.UpdateStrategy != nil {
-		if out.Spec.UpdateStrategy.RegistryPoll != nil {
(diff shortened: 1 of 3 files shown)


dinhxuanvu:
@dtfranz Please hold on the retest as this e2e-upgrade is failing for real due to some issues.

dinhxuanvu:
/hold
until e2e-upgrade is fixed (non-olm related)

dtfranz:
Hi @bandrade , thanks for checking on this. 
The test you are running has a CatalogSource with an `updateStrategy` that is `nil`, however the code change here is meant only to address an `updateStrategy` that is non-nil and contains a `nil` `registryPoll`. 

I've also confirmed from both the v0.23.1 and v0.24.0 upstream OLM releases that `updateStrategy: {}` will cause the catalog-operator pod to crash-loop. These two releases are before and after this code change respectively. I also tested this on a 4.12 cluster and encountered the same issue. It's safe to say that this is an existing issue not introduced by this code change, and I feel as though it should be addressed through another PR. 

If that's acceptable to you then I'd like to merge this change and start a follow-up to fix the problem arising from `updateStrategy: {}`. But if not it will take some time to fix the other issue and pull that fix into this PR as well.

dtfranz:
Looks like the e2e-upgrade test is passing again, gonna retest here.
/retest

dinhxuanvu:
The CI has passed.
/unhold

PR #473, opened 2023-03-15
OCPBUGS-7910: Sort channels in lexicographical order in Packagemanifest(#2925)

anik120:
See https://github.com/operator-framework/operator-registry/issues/1069 for more details

Upstream-repository: operator-lifecycle-manager
Upstream-commit: e4b15326c297f610dbf569d771abd62a703b0eb2

PR #474, opened 2023-03-15
OCPBUGS-7676: Update helm to v3.11.1 for CVE-2023-25165 

tmshort:
Also syncs everything from upstream repos as of 2023-03-15.


anik120:
@tmshort the "syncing everything from upstream" part consists of a lot of bug fixes. Are we okay with asking QE to do bulk testing using a single PR?  

Last I remember we were not doing bulk syncs since it's harder for QE to test, or much harder for us to quickly revert if we inadvertently break Openshift CI. 

Not sure if the process/messaging/thinking around those have changed recently.

tmshort:
The helm and 1.26 changes are already upstream as a single commit/merge (https://github.com/operator-framework/operator-lifecycle-manager/pull/2933), so step 1 effectively encompasses everything (hence this PR). 

Also, the CIs likely won't pass until the final step (see #465). If we're OK with that, then I can break this up, into at least 2 steps.

I could separate out the helm+1.26 changes into a separate PR after getting the rest of the upstream fixes merged.
The helm changes themselves are relatively minor (really just go.mod), so steps 2 and 3 could be combined.

tmshort:
> That way we can hand QE individual pieces to test (and diagnose/dissect individual commits if problems arise)

Hmm... we are very far behind, and @perdasilva was trying to get us ahead . Given the number of upstream commits, that's a lot of effort (it's really just getting PRs tested and approved, as demonstrated by this PR, all the code has been merged locally).

I'm concerned about the calendar time this will take, and getting the CI tests to behave.

tmshort:
> Also, if we can guarantee we'll not break anything in openshift with any of the commit, thereby requiring a full revert of all the PR later on, then we can say we can avoid any pain for ourselves now. But can we 100% guarantee something like that won't happen? :)

Understood. Would you be more comfortable with small groupings of commits (say two or three)? That will reduce the time/effort to a-half or a-third? There are ~50 (depending on how you count) commits to be downstreamed.

tmshort:
I was trying to `bisect` the PR list to figure out what broke what test, but splitting things up will make it easier to find those problems.

anik120:
> After speaking with @kevinrizza, I'll probably do it as individual or small batches (2 or 3) commits at a time. Are you good with that @anik120 ?

Yup sounds good to me, thanks Todd! 

I'm going to close https://github.com/openshift/operator-framework-olm/pull/473 with a comment in https://issues.redhat.com/browse/OCPBUGS-7910 that a batch PR will pick this up and move the ticket to post.

PR #480, opened 2023-04-06
OPRUN-2913: Downstream sync

tmshort:
Downstream sync from the operator-framework/api repo through [PR277](https://github.com/operator-framework/api/pull/277).
[PR278](https://github.com/operator-framework/api/pull/278) is a bump to k8s 1.26, so it will need to be done later.

No Jira was made for these, so I'm using the [OPRUN-2913](https://issues.redhat.com//browse/OPRUN-2913) downstreaming ticket.


tmshort:
Ugh... all tests pass and now I have to run them again :(

PR #486, opened 2023-04-27
OCPBUGS-1684: Optimize certificate generation

tmshort:
Reduce certificate generation to once-a-day.
Optimize loading of the certificate.
Rename operations to correctly identify what they do.

### DIFF
--- a/cmd/collect-profiles/main.go
+++ b/cmd/collect-profiles/main.go
@@ -32,6 +32,8 @@ import (
 const (
 	profileConfigMapLabelKey = "olm.openshift.io/pprof"
 	pprofSecretName          = "pprof-cert"
+	validity                 = 24 * time.Hour
+	minValidity              = 15 * time.Minute
 )
 
 var (
@@ -74,7 +76,7 @@ func newCmd() *cobra.Command {
 		Short: "Retrieves the pprof data from a URL and stores it in a configMap.",
 		Long: `The collect-profiles command makes https requests against pprof URLs
 		provided as arguments and stores that information in immutable configMaps.
-		
+
 		# Example command with multiple arguments
 		./collect-profiles -n - openshift-operator-lifecycle-manager \
                 - --config-mount-path \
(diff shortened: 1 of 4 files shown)


tmshort:
The first job took 17s, subsequent jobs are taking 3~4s:
```
NAMESPACE                              NAME                        COMPLETIONS   DURATION   AGE
openshift-operator-lifecycle-manager   collect-profiles-28043685   1/1           17s        38m
openshift-operator-lifecycle-manager   collect-profiles-28043700   1/1           4s         24m
openshift-operator-lifecycle-manager   collect-profiles-28043715   1/1           4s         9m37s
openshift-operator-lifecycle-manager   collect-profiles-28043730   1/1           3s         8m46s
openshift-operator-lifecycle-manager   collect-profiles-28043745   1/1           4s         114s
```
Assuming all nodes are roughly equivalent; this is a significant time savings.

cmd/collect-profiles/main.go
```diff
@@ -143,15 +145,23 @@ func newCmd() *cobra.Command {
...
 			}
 
-			httpClient, err := getHttpClient(certPath, keyPath)
-			if err != nil {
-				return err
+			httpClient := &http.Client{
+				Transport: &http.Transport{
+					TLSClientConfig: &tls.Config{
+						InsecureSkipVerify: true,
+						Certificates:       []tls.Certificate{*tlsCert},
```
m1kola, line 162: I know it was like that originally in `getHttpClient`, but this thing makes me nervous.

Do you know why we have to use `InsecureSkipVerify`? Can't we set `RootCAs` to a `x509.CertPool` containing correct cert so that our self singed cert can be verified?

tmshort: TL;DR: I think it's a bit beyond the scope of this Bug/PR to fix the certificate verification.

I was just focusing on the certificate generation and not the underlying security policy.

The pod template in the CronJob does not mount the `olm-operator-serving-cert` secret, which should be the certificate (_and_ key) of the server. So it does not have access. I don't know what the CA is for that certificate (or if it's self-signed - as I have not figured out where it's generated - I'm guessing some openshift/cert-mgr magic. @awgreene do you know?).

(Note that there are two pods from which profiling information is collected, so both secrets, the other being `catalog-operator-serving-cert`, would need to be referenced.)

The collect-profiles pod collects from two different pods. It would require changing the manifests to mount the secret(s), adding a CLI option to reference the secret(s), then code to reference. As the  it might require multiple arguments, and that doesn't scale. If I could figure out where the certificate comes from, and if the two certificates use a common CA, that would ease the burden.

I will note that the olm and catalog pods _do_ appear to validate the client, as they mount the `pprof` secret.

tmshort: Follow up: the `catalog-operator-serving-cert` and `olm-operator-serving-cert` are both signed by `CN=openshift-service-serving-signer@1683311309` (presumably the number at the end is unique per instance). The signing cert is included in the certificate chain.

This certificate (_and_ key) is located in the `signing-key` secret in the `openshift-service-ca` namespace; so no access. The key being there is a bit dangerous.

Optimally, the CA cert (_only_) would be made available _somewhere_, and I'm not sure that's the case. There are no secrets in the `default` namespace.

[Service CA certificates](https://docs.openshift.com/container-platform/4.12/security/certificate_types_descriptions/service-ca-certificates.html) describes the behavior of the OpenShift CA, and how to use it. Rotation would have to be handled (OLM does this already), and the collect-profiles pod really ought to use the same certificate generation mechanism, but it's designed for services (servers) not clients, AFAICT (a fake service could be defined, though).

Definitely beyond the scope of this bugfix/PR.

tmshort: @m1kola ?

awgreene (approved):
Thanks for all the variable name change improvements.
/approve

cmd/collect-profiles/main.go
awgreene: I'm fine with this change for now as it addresses the bug defined in the ticket. We should probably consider if we should disable this job by default as I don't believe anyone has actually utilized the pprof data collected by the job since this feature was introduced.

m1kola: > TL;DR: I think it's a bit beyond the scope of this Bug/PR to fix the certificate verification.

Absolutely. I implied that when I said it was like that in `getHttpClient`, but I'm sorry that I did not make it more clear.

m1kola (approved):
I think we should create a ticket to address `InsecureSkipVerify`. It has potential for man in the middle attack.

But I agree that it is not in the scope of this change.

/lgtm

PR #487, opened 2023-04-28
Ocpbugs 1684n

tmshort:
This removes the client certificate generation. mTLS is not used, but TLS still is.


tmshort:
Other PR merged,

PR #498, opened 2023-06-22
OPRUN-2995: Remove dependency on cluster policy controller in favor of hardcoding

perdasilva:
Due to PSA changes, we had to implement another controller in the downstream to label non-payload openshift-* namespaces with CSVs to signal security team's cluster policy controller (CPC) to manage the namespace's PSA security standard label.

Our controller depends on an ignore list exported by the CPC so that we can ignore the same payload namespaces.

There are two problems with this:
 1. OLM's must depend on CPC for the same OpenShift version i.e. OLM@release-4.12 -> CPC@release-4.12
 2. CPC is also a controller and has many of the same dependencies as us (kube api, etc.) and they can interfere with eachother possibly causing vendoring issues. Which is not the type of problem you want when importing a list of strings...

It doesn't seem like there's any good solution here. For our own QoL, this PR removes the dependency and copies the list over to our project. Of course, this can lead to skew...

Having spoken to the security team, they don't think this will change particularly often. In the meantime, we should find some other way to deal with this. Maybe reach out to the security team and ask them to split that list out into its own library and use straight up go types instead of a k8s `Set`


perdasilva:
label docs-approved qe-approved px-approved lgtm

perdasilva:
This is a bit of maintenance and shouldn't affect product. Therefore I've applied the px/docs/qe-approved labels myself. Also, the verify stage fails because of changes to the staging directory. These changes are downstream only and were necessary. It should be ok to override on these grounds.

tmshort (approved):
Seems reasonable. I'm not seeing the equivalent code upstream, which is expected?

PR #500, opened 2023-07-03
OCPBUGS-15737: Re-enable psa plugin

perdasilva:
Upstream PSA changes were introduced in Kubernetes 1.25 (OpenShift 4.12). Downstream OLM has a plug-in that integrates with the security team's Cluster-Policy-Constroller that helps smooth out the transition and helps avoid stuck workloads for our customers. We had to disable the plug-in in 4.12. It's not ready to be re-enabled.


perdasilva:
/override ci/prow/verify

this stage fails because there are changes to the staging directory. These changes affect downstream files only. ok to override.

perdasilva:
/override verify

Changes here are in downstream only files in the staging (which normally is closed for changes outside of upstream commits). This is one of the rare cases where we need to make changes to downstream only behavior that must exist in the olm src tree in the staging directory.

tmshort (approved):

PR #506, opened 2023-07-10
Registry bulk sync

grokspawn:
Current status:  cannot `make verify` because `manifests` make target is broken due to interaction with helm/k8s dependency bumps.


tmshort:
Please see #510

PR #526, opened 2023-07-31
OCPBUGS-17157: operator-registry: cherry-pick to be up-to-date

stevekuznetsov:


stevekuznetsov:
The `make` target succeeds for me locally - would appreciate some guidance here.

pkg/manifests/csv.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: packageserver
   namespace: openshift-operator-lifecycle-manager
   labels:
-    olm.version: 0.19.0
+    olm.version: 0.0.0-99ea6d7f94caf2b8bbc3b4dc579884089e06abcf
```
grokspawn, line 8: This doesn't look right, and I think it will result in a CI failure.

tmshort: `make verify` also does `make verify-manifests`, and also does a bunch of other checks

tmshort: Unfortunately, the documentation I wrote (and the scripts, darn it!) do not include this information.

tmshort:
Noting that I see `go.mod` + `vendor` changes, but not `go.sum`; presumably since **ci/prow/verify** passed, the checksum was already there.
/lgtm

PR #531, opened 2023-08-03
OCPBUGS-17157: sync

stevekuznetsov:


stevekuznetsov:
It depends on what the server does with unknown fields on the CatalogSource API - it's backwards compatible so likely the fields will just be ignored on versions of the server that don't understand them. That being said, while it's likely safe to merge before, no need to do it.

stevekuznetsov:
@joelanford I think your comment (and my response) are actually for https://github.com/operator-framework/operator-marketplace/pull/527

tmshort (approved):
/lgtm

PR #533, opened 2023-08-07
fix the manifests generation

stevekuznetsov:
manifests: use the CVO-replaced placeholder for version

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

makefile: verify that we use the CVO-based placeholder

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

PR #534, opened 2023-08-07
OCPBUGS-17157: scripts: add a Go-based bumper, sync upstream

stevekuznetsov:
scripts: add a Go-based bumper

This script sohuld be easier to maintain, automate, and have less
historical baggage as compared to the shell scripts. As an example, the
problem space of "cherry-pick some small number of commits on top of an
in-sync branch" is much smaller than the "cherry-pick some unknownn
number of commits on some dirty branch state with out-of-order commits
already present."

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -86,9 +86,12 @@ or amended to the last commit of the branch.
 
 Once `make -k verify` is resolved, create a PR from this working sync branch.
 
-## TODO
+# Long-lived Carry Commits
 
-* Add `make verify` to the `sync_pop_candidate.sh` and/or `sync.sh` scripts.
+It is required at times to write commits that will live in the `vendor/` directory
+on top of upstream code and for those commits to be carried on top for the forseeable
+future. In these cases, prefix your commit message with `[CARRY]` to pass the commit
+verification routines.
 
 ## References
 1. [Downstream to operator-framework-olm](https://spaces.redhat.com/display/OOLM/Downstream+to+operator-framework-olm)
--- a/scripts/bumper/main.go
+++ b/scripts/bumper/main.go
@@ -277,7 +277,7 @@ func cherryPick(ctx context.Context, logger *logrus.Entry, c commit) error {
 			"go", "mod", "verify",
 		), os.Environ()...),
 		withEnv(exec.CommandContext(ctx,
-			"make", "manifests", "OLM_VERSION=0.0.1-snapshot",
+			"make", "generate-manifests",
 		), os.Environ()...),
 		exec.CommandContext(ctx,
 			"git", "add",
--- a/scripts/bumper/main.go
+++ b/scripts/bumper/main.go
@@ -206,7 +206,7 @@ func detectNewCommits(ctx context.Context, logger *logrus.Entry, stagingDir stri
 				infoCmd := exec.CommandContext(ctx,
 					"git", "show",
 					line,
-					"--pretty=format:%H|%cI|%an|%s",
+					"--pretty=format:%H\u00A0%cI\u00A0%an\u00A0%s",
 					"--quiet",
 				)
 				stdout, stderr := bytes.Buffer{}, bytes.Buffer{}
@@ -216,7 +216,7 @@ func detectNewCommits(ctx context.Context, logger *logrus.Entry, stagingDir stri
 				if err := infoCmd.Run(); err != nil {
 					return nil, fmt.Errorf("failed to run command: %s %s: %w", stdout.String(), stderr.String(), err)
 				}
-				parts := strings.Split(stdout.String(), "|")
+				parts := strings.Split(stdout.String(), "\u00A0")
 				if len(parts) != 4 {
 					return nil, fmt.Errorf("incorrect parts from git output: %v", stdout.String())
 				}
--- a/scripts/bumper/main.go
+++ b/scripts/bumper/main.go
@@ -1,0 +1,320 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/fs"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+	"text/tabwriter"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+type mode string
+
+const (
+	summarize   mode = "summarize"
+	synchronize mode = "synchronize"
+)
+
+type options struct {
+	stagingDir       string
+	commitFileOutput string
+	commitFileInput  string
+	mode             string
+	logLevel         string
+}
+
+func (o *options) Bind(fs *flag.FlagSet) {
+	fs.StringVar(&o.stagingDir, "staging-dir", "staging/", "Directory for staging repositories.")
+	fs.StringVar(&o.mode, "mode", string(summarize), "Operation mode.")
+	fs.StringVar(&o.commitFileOutput, "commits-output", "", "File to write commits data to after resolving what needs to be synced.")
+	fs.StringVar(&o.commitFileInput, "commits-input", "", "File to read commits data from in order to drive sync process.")
+	fs.StringVar(&o.logLevel, "log-level", logrus.InfoLevel.String(), "Logging level.")
+}
+
+func (o *options) Validate() error {
+	switch mode(o.mode) {
+	case summarize, synchronize:
+	default:
+		return fmt.Errorf("--mode must be one of %v", []mode{summarize, synchronize})
+	}
+
+	if _, err := logrus.ParseLevel(o.logLevel); err != nil {
+		return fmt.Errorf("--log-level invalid: %w", err)
+	}
+	return nil
+}
+
+func main() {
+	logger := logrus.New()
+	opts := options{}
+	opts.Bind(flag.CommandLine)
+	flag.Parse()
+
+	if err := opts.Validate(); err != nil {
+		logger.WithError(err).Fatal("invalid options")
+	}
+
+	logLevel, _ := logrus.ParseLevel(opts.logLevel)
+	logger.SetLevel(logLevel)
+
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer cancel()
+
+	var commits []commit
+	var err error
+	if opts.commitFileInput != "" {
+		rawCommits, err := os.ReadFile(opts.commitFileInput)
+		if err != nil {
+			logrus.WithError(err).Fatal("could not read input file")
+		}
+		if err := json.Unmarshal(rawCommits, &commits); err != nil {
+			logrus.WithError(err).Fatal("could not unmarshal input commits")
+		}
+	} else {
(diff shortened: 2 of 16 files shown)


pkg/manifests/csv.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: packageserver
   namespace: openshift-operator-lifecycle-manager
   labels:
-    olm.version: 0.19.0
+    olm.version: 0.0.0-6b8621226ef93b41e980e1c35f843e5c97f145a8
```
grokspawn, line 8: I really don't like the behavior around this tag, but not using a constant tag breaks verify: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_operator-framework-olm/534/pull-ci-openshift-operator-framework-olm-master-verify/1688556770030522368

stevekuznetsov: @grokspawn yep! The linked dependent PR (https://github.com/openshift/operator-framework-olm/pull/533) fixes that and should merge first, after which I'll rebase this

scripts/bumper/main.go
joelanford, line 1: Do/will we have docs about when/how to use this script? Will there be a follow-up PR to remove the obsolete shell scripts? Just trying to understand the big picture.

stevekuznetsov: Yes, as this improves I'll add docs and remove the old bits. I'll be the one doing syncs for the near term while the bugs get ironed out.

stevekuznetsov:
Needs https://github.com/operator-framework/operator-lifecycle-manager/pull/3007

stevekuznetsov:
Rebased, should be ready

stevekuznetsov:
Can't reproduce the unit failure locally...

```
$ go test -count 1 ./pkg/controller/operators/olm/plugins/...
ok  	github.com/operator-framework/operator-lifecycle-manager/pkg/controller/operators/olm/plugins	1.715s
```

stevekuznetsov:
Verify job is failing since there's no mechanism in this repo to mark commits as expected downstream carries. I will add such a mechanism when I'm back from PTO on Monday.

Unit job is failing on waiting for informers to sync  from a fake client, but this is not something I can reproduce myself locally. Will need to look into it when I'm back.

tmshort:
Missing one at least one upstream commit... but it's just OWNERS update, which doesn't really matter.

scripts/verify_commits.sh
```diff
@@ -37,9 +37,12 @@ function verify_downstream_only() {
     local inside_staging
     inside_staging="$(git show --name-only "${downstream_commit}" -- staging)"
     if [[ -n "${inside_staging}" ]]; then
-        err "downstream non-staging commit ${downstream_commit} changes staging"
+        if git log -n 1 "${downstream_commit}" --pretty=%s | grep -q '[CARRY]'; then
```
tmshort, line 40: Why `[CARRY]`?

scripts/bumper/main.go
```diff
@@ -0,0 +1,320 @@
...
+				)); err != nil {
+					return err
+				}
+			} else {
+				return err
+			}
+		}
+	}
+
+	for _, cmd := range []*exec.Cmd{
```
tmshort, line 269: A bit concerned here about the difference between this and `scripts/sync_pop_candidates.sh` (plus `sync.sh`). Some of it has to do with direct invocation of some of these commands, which might be better suited to do via `make`.

The sync scripts do:

* `make vendor` (which does tidy/vendor/verify)
* `make verify-manifests` (which does `make manifests` with `OLM_VERSION` set, which I think is a better way to do it)
*  `make verify-nested-vendor` (which may not be necessary)
*  `make verify-commits` (which is not done here, and probably should be "just in case", at least initially)

This lacks the recovery mechanisms in the the scripts, but those may not be needed if this is run frequently enough.

If this is to replace the sync script, then the sync scripts should be modified to build and call this instead.

I'm also not seeing a new branch created to run this in; does it do it on the local `master` branch?

scripts/bumper/main.go
```diff
@@ -0,0 +1,320 @@
...
+			return nil, err
+		}
+
+		for _, line := range strings.Split(output, "\n") {
+			line = strings.TrimSpace(line)
+			if line != "" {
+				infoCmd := exec.CommandContext(ctx,
+					"git", "show",
+					line,
+					"--pretty=format:%H|%cI|%an|%s",
```
tmshort, line 209: A bit concerned here... mostly because I could see a `|` being used in the message, and the handling of multi-line messages.

scripts/bumper/main.go
```diff
@@ -0,0 +1,320 @@
...
+		logger.Debug("detecting commits")
+		output, err := runCommand(logger, exec.CommandContext(ctx,
+			"git", "log",
+			"-n", "1",
+			"--grep", "Upstream-repository: "+path,
+			"--grep", "Upstream-commit",
+			"--all-match",
+			"--pretty=%B",
+			"--",
+			filepath.Join(stagingDir, path),
```
tmshort, line 159: https://github.com/go-git/go-git ?

tmshort (changes requested):
Some concerns around the bumper; the other stuff looks fine.

scripts/bumper/main.go
tmshort: `go-get` might offer better interface to go, particularly the output.

scripts/verify_commits.sh
stevekuznetsov: I don't have a strong opinion on this. Carry is the word OCP uses for commits that are carried on top of upstream for the forseeable future.

scripts/bumper/main.go
stevekuznetsov: > Some of it has to do with direct invocation of some of these commands, which might be better suited to do via make.

I was getting weird segfaults doing that and it's a minimal risk I don't really think matters.

> The sync scripts do:

I don't want this to do that. This syncs. If you want to verify, you can do that yourself. An automated syncer does the syncing, and pushes that up to get CI. If CI fails, we see it.

> If this is to replace the sync script

Give me some time. This will be improved and made automatic soon enough.

> I'm also not seeing a new branch created

Again, I don't want it to. If you want a new branch, do that. This tool works in the current context. The libraries that I'll tap into for publishing this as a PR and all that expect as much. I also (personally) prefer controlling this externally.

scripts/bumper/main.go
stevekuznetsov: Changed to use a non-breaking space

scripts/bumper/main.go
tmshort: I'm concerned about the `git show --pretty` output and parsing of it.

scripts/bumper/main.go
stevekuznetsov: Consider that if we run the `make verify*` bits, we will need to write code here to disambiguate between "we failed to generate new manifests" and "the verification of manifests failed, since we're just generating them for the first time and expect to have a git diff". Not really clear to me how that is better than running `make manifests` directly.

scripts/bumper/main.go
tmshort: The format used in the sync scripts uses space for separators, but the individual components won't include spaces.

scripts/bumper/main.go
stevekuznetsov: > sync scripts should be updated to use this (if this goes in)

My intent is to deliver a series of small PRs that build this up into the fully-fledged codebase it will become. I hope we can accommodate this ... ? I was using this proto-bumper already and figured it could stand alone while the rest got implemented.

scripts/bumper/main.go
stevekuznetsov: What's the specific concern with this command? I pass `"--pretty=%B"` and there's no parsing here.

tmshort: Not this parsing, but the other one... sorry... slightly different view.

stevekuznetsov: I amended the other one to use a non-breaking space as the delimiter. I am happy to fix further when automated syncing fails to work after someone uses a non-breaking space in their git commit message.

scripts/bumper/main.go
stevekuznetsov: To be clear, the risk here is of a pretty small magnitude. If `make verify-manifests` ever changes what it expects, and we did not remember to update this script, the next automated sync PR will fail tests. I'll add a `bump-manifests` target since that will be useful to individual devs as well.

tmshort: Which is why I'm suggesting `bump-manifests`

scripts/bumper/main.go
```diff
@@ -0,0 +1,320 @@
...
+			return nil, err
+		}
+
+		for _, line := range strings.Split(output, "\n") {
+			line = strings.TrimSpace(line)
+			if line != "" {
+				infoCmd := exec.CommandContext(ctx,
+					"git", "show",
+					line,
+					"--pretty=format:%H\u00A0%cI\u00A0%an\u00A0%s",
```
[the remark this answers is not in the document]
stevekuznetsov: Not sure if you meant that literally but `\u00A0` is the Unicode non-breaking space and is distinct from the new-line. Chances of someone putting that into their message are slim.

scripts/bumper/main.go
tmshort: The likelyhood of that is fairly low. I _was_ about to suggest that the author and subject aren't necessary for the processing, but I see you're using them for reporting.

scripts/verify_commits.sh
tmshort: I'm ambivalent, but this ought to be documented somewhere.

stevekuznetsov: Where would you suggest? As it stands the failure messages from the script are fairly explicit - if you hit this case by writing a commit that you mean to carry, you will see error messages telling you how to change the commit message. Happy to add it to some other location you think folks will be likely to read.

scripts/bumper/main.go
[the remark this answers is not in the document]
tmshort: At first I wrote newline, but then realized it wasn't a newline! But I agree that it's very unlikely someone would use that character.

scripts/verify_commits.sh
tmshort: I have no suggestions for an alternative; but mostly just document that commits with "whatever-prefix-is-finally-chosen" in the headline are treated special. Either in the `README.md` in the `scripts` directory or in Confluence.

scripts/README.md
```diff
@@ -86,9 +86,12 @@ or amended to the last commit of the branch.
 
 Once `make -k verify` is resolved, create a PR from this working sync branch.
 
-## TODO
+# Long-lived Carry Commits
 
-* Add `make verify` to the `sync_pop_candidate.sh` and/or `sync.sh` scripts.
```
tmshort, line 91: Thanks for removing this.

tmshort (approved):
/lgtm

The sync scripts should be updated once we're confident in these changes.

PR #537, opened 2023-08-10
OPRUN-3022: Add support for make verify to sync script

tmshort:
This updates the sync script to use the same targets as "make verify" This minimizes the repetative nature of calling "make verify", and avoids the incorrect update of csv.yaml by "make manifests"

Also prefix the Makefile "echo" commands with an @


ankitathomas (approved):
/lgtm

PR #540, opened 2023-08-14
OCPBUGS-17791: opm: always serve pprof endpoints, improve server allocations (#1129)

stevekuznetsov:
* pkg/cache: use a shared buffer to limit allocations

Previously, new buffers were allocated on each file we read in, which was unnecessary and wasteful.

* cmd/opm: serve pprof endpoints by default

There is no substantial runtime cost to serving pprof endpoints, and when things hit the fan and we need to investigate performance in situ, there is no time to restart pods and change flags. Capturing profiles remains opt-in, since those are costly.

---------

Upstream-commit: 68e13df96590977370ffcd1a8e9ff76e0f2a03f2

### DIFF
--- a/staging/operator-registry/cmd/opm/serve/serve.go
+++ b/staging/operator-registry/cmd/opm/serve/serve.go
@@ -36,8 +36,9 @@ type serve struct {
 	port           string
 	terminationLog string
 
-	debug     bool
-	pprofAddr string
+	debug           bool
+	pprofAddr       string
+	captureProfiles bool
 
 	logger *logrus.Entry
 }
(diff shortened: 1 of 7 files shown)


bparees:
can you link to the 4.14 version of this PR so we can confirm it went into 4.14 before backporting to 4.13?

stevekuznetsov:
@bparees sure thing! The other PR was https://github.com/openshift/operator-framework-olm/pull/526

KeenonLee (commented):
/label qe-approved
/label cherry-pick-approved

PR #550, opened 2023-08-17
scripts/bumper: automate pushing changes and creating PRs

stevekuznetsov:


scripts/bumper/main.go
```diff
@@ -99,10 +166,39 @@ func main() {
...
+	for _, commit := range commits {
+		commitLogger := logger.WithField("commit", commit.Hash)
+		missing, err := isCommitMissing(ctx, commitLogger, opts.stagingDir, commit)
+		if err != nil {
+			commitLogger.WithError(err).Fatal("failed to determine if commit is missing")
+		}
+		if missing {
+			missingCommits = append(missingCommits, commit)
+		}
+	}
```
tmshort, line 179: Wondering why this is necessary given `detectNewCommits()`? This isn't moved from elsewhere like `cherryPickAll` below.

stevekuznetsov: While testing, I found it really useful to keep track of all the commits that need to be cherry-picked, even if some have already been picked. I added the `centralRef` flag to mark the start of our search, and this check to ensure we don't try to cherry-pick a commit more than once. This is not as necessary for the cherry-picks themselves, but was crucial for making sure the table of commits in the PR we send is full, even if the last invocation of this tool that pushes the PR did not do every single cherry-pick.

tmshort:
/retest
(Is this the definition of insanity?)

stevekuznetsov:
Yes, let's override it and move on - there's no world in which the content of this PR can influence that test...

stevekuznetsov:
Speak of the devil...

ncdc (approved):

PR #555, opened 2023-08-25
 OCPBUGS-17950: Make packageserver wakeup interval configurable

tmshort:
The packageserver wakeup interval is used to update its package manifest from CatSrcs. However, this does not need to happen periodically (especially 5m) as the CatSrc pods will restart when a change occurs to data it is caching, and this will be acted upon by packageserver.

This allows configuration of this interval via the package-server-manager deployment (which then gets passed onto the packageserver deployment).


tmshort:
Alternative to #553

PR #556, opened 2023-08-29
OCPBUGS-17041: Release Leader Election on Manager Exit

awgreene:
This commit introduces a couple of changes to the package server manager
to improve hand off between running pods during an upgrade or
redeployment.

1. The package server manager will now voluntarily release its lease on
manager exit which will speed up voluntary leader transition as the new
leader shouldn't have to wait the LeaseDuration time before taking
over.

I should note that enabling this setting expects that the binary will
immediately exit upon release.

2. The package server manager deployment has had its .strategy.type
field updated from Rolling to Recreate, which will prevent the new pod
from attempting to acquire the lease before the current leader has
shutdown and released its lease.

### DIFF
--- a/cmd/package-server-manager/main.go
+++ b/cmd/package-server-manager/main.go
@@ -80,17 +80,18 @@ func run(cmd *cobra.Command, args []string) error {
 
 	packageserverCSVFields := fields.Set{"metadata.name": name}
 	mgr, err := ctrl.NewManager(restConfig, manager.Options{
-		Scheme:                  setupScheme(),
-		Namespace:               namespace,
-		MetricsBindAddress:      metricsAddr,
-		LeaderElection:          !disableLeaderElection,
-		LeaderElectionNamespace: namespace,
-		LeaderElectionID:        leaderElectionConfigmapName,
-		LeaseDuration:           &le.LeaseDuration.Duration,
-		RenewDeadline:           &le.RenewDeadline.Duration,
-		RetryPeriod:             &le.RetryPeriod.Duration,
-		HealthProbeBindAddress:  healthCheckAddr,
-		PprofBindAddress:        pprofAddr,
+		Scheme:                        setupScheme(),
+		Namespace:                     namespace,
+		MetricsBindAddress:            metricsAddr,
+		LeaderElection:                !disableLeaderElection,
+		LeaderElectionNamespace:       namespace,
+		LeaderElectionID:              leaderElectionConfigmapName,
+		LeaseDuration:                 &le.LeaseDuration.Duration,
+		RenewDeadline:                 &le.RenewDeadline.Duration,
+		RetryPeriod:                   &le.RetryPeriod.Duration,
+		HealthProbeBindAddress:        healthCheckAddr,
+		PprofBindAddress:              pprofAddr,
+		LeaderElectionReleaseOnCancel: true,
 		Cache: cache.Options{
 			ByObject: map[client.Object]cache.ByObject{
 				&olmv1alpha1.ClusterServiceVersion{}: {
--- a/manifests/0000_50_olm_06-psm-operator.deployment.ibm-cloud-managed.yaml
+++ b/manifests/0000_50_olm_06-psm-operator.deployment.ibm-cloud-managed.yaml
@@ -9,7 +9,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
 spec:
   strategy:
-    type: RollingUpdate
+    type: Recreate
   replicas: 1
   selector:
     matchLabels:
--- a/scripts/generate_crds_manifests.sh
+++ b/scripts/generate_crds_manifests.sh
@@ -123,7 +123,7 @@ metadata:
   annotations:
 spec:
   strategy:
-    type: RollingUpdate
+    type: Recreate
   replicas: 1
   selector:
     matchLabels:
--- a/manifests/0000_50_olm_06-psm-operator.deployment.yaml
+++ b/manifests/0000_50_olm_06-psm-operator.deployment.yaml
@@ -9,7 +9,7 @@ metadata:
     include.release.openshift.io/self-managed-high-availability: "true"
 spec:
   strategy:
-    type: RollingUpdate
+    type: Recreate
   replicas: 1
   selector:
     matchLabels:


tmshort:
/lgtm
pending successful CI completion

cmd/package-server-manager/main.go
```diff
@@ -71,17 +71,18 @@ func run(cmd *cobra.Command, args []string) error {
...
-		RetryPeriod:             &le.RetryPeriod.Duration,
-		HealthProbeBindAddress:  healthCheckAddr,
-		PprofBindAddress:        pprofAddr,
+		Scheme:                        setupScheme(),
+		Namespace:                     namespace,
+		MetricsBindAddress:            defaultMetricsPort,
+		LeaderElection:                !disableLeaderElection,
+		LeaderElectionNamespace:       namespace,
+		LeaderElectionID:              leaderElectionConfigmapName,
+		LeaderElectionReleaseOnCancel: true,
```
awgreene, line 80: This is the bulk of the change in the PR. The Go docs describing the field can be found [here](https://github.com/alvaroaleman/controller-runtime/blob/222fb669e109f4bffdf7fc14a2ac02a2a6e496ba/pkg/manager/manager.go#L190-L195), and the PR introducing the functionality can be found [here](https://github.com/kubernetes-sigs/controller-runtime/pull/1126/files#diff-d500fbd6a2aa620607ca5e2a7c3ac4f1a4c82309d1a549561e92abfcb18f2f0e).

tmshort: This will actually make #555 work faster; right now, one has to wait 2-3 minutes for the change to propagate due to the lease used in the leader election.

manifests/0000_50_olm_06-psm-operator.deployment.ibm-cloud-managed.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
 spec:
   strategy:
-    type: RollingUpdate
+    type: Recreate
```
awgreene, line 12: @tmshort this will slow down the new pod from coming up, but will prevent the new pod from attempting to acquire a lease before the other pod has shut down, meaning that it won't wait the retry duration, which should yield overall speed improvements.

scripts/generate_crds_manifests.sh
```diff
@@ -119,7 +119,7 @@ metadata:
   annotations:
 spec:
   strategy:
-    type: RollingUpdate
+    type: Recreate
```
awgreene, line 122: @tmshort this will slow down the new pod from coming up, but will prevent the new pod from attempting to acquire a lease before the other pod has shut down, meaning that it won't wait the retry duration, which should yield overall speed improvements.

awgreene:
Turns out this is a must have in 4.14 for telco, I've been asked to merge it into 4.15 and then 4.14.
/label acknowledge-critical-fixes-only

ncdc:
Removed the critical fix label. This is not a critical fix. We'll wait for the branch to open for normal development before merging.

wking:
[This][1]:

```
	Error:      	Received unexpected error:
	            	timed out waiting for the condition
	Test:       	Operator Group CSV copy watching all namespaces
```

seems unrelated.

/test e2e-gcp-olm

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_operator-framework-olm/556/pull-ci-openshift-operator-framework-olm-master-e2e-gcp-olm/1704209668647161856

wking:
[PR test history][1] seems like something more fundamental than random-flakes is going on with this job:

image

But [history for that job][2] does include a few [recent][3] [passes][4].  But `timed out waiting for the condition`  isn't a particularly transparent error condition.  Maybe the test suite could be updated to also dump the current state of the condition being waited for and explain how its different from what the suite was expecting?

[1]: https://prow.ci.openshift.org/pr-history/?org=openshift&repo=operator-framework-olm&pr=556
[2]: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-operator-framework-olm-master-e2e-gcp-olm
[3]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_operator-framework-olm/563/pull-ci-openshift-operator-framework-olm-master-e2e-gcp-olm/1703760810888663040
[4]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_operator-framework-olm/551/pull-ci-openshift-operator-framework-olm-master-e2e-gcp-olm/1704465258321547264

ncdc:
@wking 💯 we are working on improving test debuggability

ncdc:
Looking into the latest e2e flake/failure.

1. e2e started polling for a succeeded CSV at 15:48:34.396258Z
2. scheduler assigned webhook pod to a node at 15:49:31.218077. Note this is about 3 seconds before the 1-minute polling expires.
3. pod starts logging at 15:49:39.065Z, after polling expired
4. e2e test teardown deletes the namespace
5. pod tries to grab leader lease 15:49:39.259582 `error initially creating leader election record: configmaps "78ab7849.operators.coreos.io" is forbidden: unable to create new content in namespace webhook-e2e-gnlxr because it is being terminated`

awgreene:
The failed test doesn't seem to be related to the changes introduced in this PR based on the [periodic test grid results](https://testgrid.k8s.io/redhat-openshift-olm#periodic-ci-openshift-operator-framework-olm-master-periodics-e2e-gcp-olm), I'll take a swing at improving that test before attempting to retest this again.

ncdc:
👍 I agree it isn't related. Thanks for looking into test improvements.

tmshort:
I swear I've seen a similar failure in other tests.

ncdc:
Updated timings with more detail. I think we may need to extend the polling to more than 1 minute, or we need to start polling later, after the catalog is happy. There's just not enough time to do what we need within 1 minute when we start the polling this early:

1. `15:48:34.396258Z Start polling`
2. `time="2023-10-24T15:48:35Z" level=warning msg="error getting bundle stream" action="refresh cache" err="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 172.30.207.167:50051: i/o timeout\"" source="{catalog-gxmmh webhook-e2e-gnlxr}"`
3. `time="2023-10-24T15:49:26Z" level=info msg="added to bundle, Kind=ClusterServiceVersion" configmap=webhook-e2e-gnlxr/c2b1b338f5c023731fba9484eb4b81d35f81414ced29d9fcd90813ba9a8e8b8 key=webhook-operator.clusterserviceversion.yaml`
4. `2023-10-24T15:49:31.156721Z deployment created`

tmshort:
Some of those errors look to be networking-type errors (i.e. `i/o timeout`), rather than kubernetes errors.

ncdc:
You are correct; however, they are errors that prevent the parts of the test from progressing and demonstrate the need for either more than a 1m poll timeout, or a way to start polling later.

tmshort:
> You are correct; however, they are errors that prevent the parts of the test from progressing and demonstrate the need for either more than a 1m poll timeout, or a way to start polling later.

Or retries, if they are not present.

tmshort:
> Updated timings with more detail. I think we may need to extend the polling to more than 1 minute, or we need to start polling later, after the catalog is happy. There's just not enough time to do what we need within 1 minute when we start the polling this early

So, the catalog _should_ be happy; in this location (webhook_e2e_test.go:685), there is `fetchCatalogSourceOnStatus(..., catalogSourceRegistryPodSynced())`. `fetchCatalogSourceOnStatus()` polls for "5m" until the `catalogSourceRegistryPodSynced()` function returns true; since the error is in not occurring in these functions, we have to assume the catalog is ready.

The polling timeout (error) is occurring later in `awaitCSV()` which is logging the progression of the CSV status in the poll loop, but we aren't reaching the desired state. Also, the nature of the `Eventually()` loop here means we'll never see the "never got correct status" message when the function attempts to end, as `Eventually()` will terminate the function early.  The `fetchCSV()` and `awaitCSV()` functions are functionally equivalent here, but have different logging output.

One thing we aren't doing is waiting for the Subscription to be created via `createSubscriptionForCatalog()`, instead, we are going immediately to `awaitCSV`; so I think we need to wait for the subscription to get to a good state before waiting for the CSV.

So, I have already increased the timeout period of the `awaitCSV()` function, but I think we need to call `fetchSubscription()` to make sure the subscription is created and in the correct state. This will both delay the wait for the CSV, as well as increase the timeout.

awgreene:
Separate failure, we'll need to look into this as well:

```
End-to-end: [It] Install Plan update catalog for subscription AmplifyPermissions

{Timed out after 60.001s.
Expected
    <bool>: false
to be true failed [FAILED] Timed out after 60.001s.
Expected
    <bool>: false
to be true
In [It] at: /go/src/github.com/openshift/operator-framework-olm/staging/operator-lifecycle-manager/test/e2e/util.go:582 @ 10/27/23 15:57:21.264
}
```

tmshort:
It's a failure during cleanup after the test completes (it's literally a `defer`). 
That being said, this test code is a mish-mash of Gomega/Ginkgo and assert/require, and this particular test doesn't have `BeforeEach`/`AfterEach` for individual test plans, so the behavior may be a bit wonky cleanup-wise, it might be trying to clean up _a lot_, especially as the function is ~3800 lines.

awgreene:
https://github.com/openshift/operator-framework-olm/pull/600 merged, which addresses e2e failures.

/retest

m1kola (commented):
From what I read about `LeaderElectionReleaseOnCancel` - this PR looks reasonable. But I'm a bit hesitant to lgtm it because I'm not well versed in this area.

PR #565, opened 2023-09-18
OCPVE-674: feat: add annotations for capabilities feature

eggfoobar:
adding annotation for capabilities to all manifest resources, this will allow users to make OLM an optional install

/hold 

This is on hold until https://github.com/openshift/api/pull/1589 is merged in.

### DIFF
--- a/manifests/0000_50_olm_00-catalogsources.crd.yaml
+++ b/manifests/0000_50_olm_00-catalogsources.crd.yaml
@@ -5,6 +5,7 @@ metadata:
     controller-gen.kubebuilder.io/version: v0.8.0
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
+    capability.openshift.io/name: "OperatorLifecycleManager"
   creationTimestamp: null
   name: catalogsources.operators.coreos.com
 spec:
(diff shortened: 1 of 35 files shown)


stevekuznetsov:
@eggfoobar I think you'll need to go upstream and add these unless you can post-process them into all manifests in the generator script

eggfoobar:
Thanks @stevekuznetsov, I wasn't able to find if we add the OCP specific annotations in the upstream projects, except in the `generate_crds_manifests.sh` in this repo. That's the script [scripts/generate_crds_manifests.sh](https://github.com/openshift/operator-framework-olm/pull/565/files#diff-e8033c7c4bf9c1999d5b66faebd942551d51544ef15dd0d43f7efca074981684) I modified, is there something similar in the upstream?

stevekuznetsov:
Oops, sorry - missed the `yq` in the loop there. I think this should be fine.

eggfoobar:
/unhold 

API Change merged in 🎉

ncdc (approved):

tmshort (approved):
/lgtm

kevinrizza:
looks like there was a hiccup with cluster initialization on all those tests

/retest

eggfoobar:
This will fail by design until the CVO change is in @tmshort, so this can be left on hold until we're done updating some preliminary tests, I'll unhold this once that validation is done and we can allow the normal process to take shape here.

manifests/image-references
```diff
@@ -1,5 +1,8 @@
 kind: ImageStream
 apiVersion: image.openshift.io/v1
+metadata:
+  annotations:
+    capability.openshift.io/name: OperatorLifecycleManager
```
wking, line 5: `image-references` doesn't need this annotation; it's consumed by `oc adm release new ...`, which is not aware of `annotations`, just to perform placeholder image replacement.

eggfoobar: Oh that's good to know, updated the resource, should be good now

eggfoobar:
/unhold

We should be all set with this now, CVO has been merged in and latest CI payloads are pulling in the change.

PR #568, opened 2023-09-25
OCPBUGS-17950: Use configmap to specify interval

tmshort:
Use a configmap `packageserver-env` with key `PACKAGESERVER_INTERVAL`
to specify a value for the package server interval (default: 5m)

Avoids having to disable (part of CVO) to update the interval.
e.g.
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: packageserver-env
data:
  PACKAGESERVER_INTERVAL: 20m
```

### DIFF
--- a/cmd/package-server-manager/main.go
+++ b/cmd/package-server-manager/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"time"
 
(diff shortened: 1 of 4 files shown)


tmshort:
Closing in favor of using `OLMConfig`

PR #570, opened 2023-09-27
OCPBUGS-17950: Use OLMconfig to specify package server interval

tmshort:

PR #572, opened 2023-10-03
Skuznets/bump upstream

stevekuznetsov:


tmshort:
@dinhxuanvu Hmmm...  It passes upstream, and I had my stuff downstreamed for testing in #570 and #571, where those changes successfully passed the e2es. So, not sure what's happening. I _did_ see some of the original e2e's failing because we couldn't deploy on EC2.

tmshort:
@dinhxuanvu I think I know what's going on... and it's due to the combination of #570 and #571 not being tested _together_.

Downstream sets a value via the CLI, which effectively changes the default value; and that isn't handled properly.

dinhxuanvu:
> @dinhxuanvu I think I know what's going on... and it's due to the combination of #570 and #571 not being tested _together_.
> 
> Downstream sets a value via the CLI, which effectively changes the default value; and that isn't handled properly.

I see. Just a reminder is that ClusterOperator API is a downstream API (OpenShift specific) so this type of problems usually won’t surface on upstream.

dinhxuanvu:
Let put a hold here until Todd resolves this issue fully.
/hold

tmshort:
>I see. Just a reminder is that ClusterOperator status is a downstream API (OpenShift specific) so this type of problems usually won’t surface on upstream.

It was packageserver, which is an upstream component, that's crash looping.

dinhxuanvu:
> > I see. Just a reminder is that ClusterOperator status is a downstream API (OpenShift specific) so this type of problems usually won’t surface on upstream.
> 
> It was packageserver, which is an upstream component, that's crash looping.

Yes the component in trouble was packageserver but the error that blocked the test was from ClusterOperator API which reported packageserver was not ready.

dinhxuanvu:
> But if packageserver wasn't crashing in the first place....
> 
> But, packageserver has been fixed upstream, so @stevekuznetsov should re-run the script.

Haha. I wasn't contradicting you on the root cause, only on the way the error surfaced up.
Anyway, the issue should be resolved now so rebase this PR and continue.

tmshort:
See #573

PR #578, opened 2023-10-09
Sync script: handle multi commit merges

tmshort:
Fix the sync script to handle commits in a single merge.
The sorting doesn't work as the commit time is all the same, and the next sortable field is the headline.
The commit time is the same, but the authorship time for each commit might be different or out of order.
Add an ordinal to the list of commits from each branch. This is used to sort commits (in original order) with the same commit time.

This is for the bash-based sync scripts.

### DIFF
--- a/scripts/sync.sh
+++ b/scripts/sync.sh
@@ -29,7 +29,7 @@ fetch_remote() {
 
 new_candidate_branch() {
     echo "Creating a sync branch if it doesn't already exist"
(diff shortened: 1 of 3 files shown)


tmshort:
This modifies no OLM code; so everything ought to pass.

grokspawn (approved):

grokspawn:
/label docs-approved
/label px-approved
/label qe-approved

downstream-only sync infrastructure changes only

PR #580, opened 2023-10-11
scripts/bumper: improve sort order

stevekuznetsov:
While we want to use date to order commits between repositories, where there is no real order, we must not change the order of commits within a repo.

### DIFF
--- a/scripts/bumper/main.go
+++ b/scripts/bumper/main.go
@@ -12,7 +12,6 @@ import (
 	"os/signal"
 	"path/filepath"
 	"regexp"
-	"sort"
 	"strings"
 	"text/tabwriter"
 	"time"
@@ -315,7 +314,7 @@ func detectNewCommits(ctx context.Context, logger *logrus.Entry, stagingDir, cen
 		return nil, fmt.Errorf("failed to walk %s: %w", stagingDir, err)
 	}
 
-	var commits []commit
+	commits := map[string][]commit{}
 	for repo, lastCommit := range lastCommits {
 		var remote string
 		switch mode {
@@ -365,7 +364,10 @@ func detectNewCommits(ctx context.Context, logger *logrus.Entry, stagingDir, cen
 				if err != nil {
 					return nil, fmt.Errorf("invalid time %s: %w", parts[1], err)
 				}
-				commits = append(commits, commit{
+				if _, ok := commits[repo]; !ok {
+					commits[repo] = []commit{}
+				}
+				commits[repo] = append(commits[repo], commit{
 					Hash:    parts[0],
 					Date:    committedTime,
 					Author:  parts[2],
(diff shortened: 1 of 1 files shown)


scripts/bumper/main.go
```diff
@@ -199,7 +201,7 @@ func main() {
 		}
 		for i, commit := range missingCommits {
 			commitLogger := logger.WithField("commit", commit.Hash)
-			commitLogger.Infof("cherry-picking commit %d/%d", i+1, len(missingCommits))
+			commitLogger.Infof("cherry-picking commit %d/%d", i+1, len(missingCommits)+1)
```
tmshort, line 204: `+1` on the length of `missingCommits`? Is that right?

scripts/bumper/main.go
```diff
@@ -284,11 +286,12 @@ func detectNewCommits(ctx context.Context, logger *logrus.Entry, stagingDir, cen
 		output, err := runCommand(logger, exec.CommandContext(ctx,
 			"git", "log",
 			centralRef,
-			"-n", "1",
+			"-n", strconv.Itoa(history),
```
tmshort, line 289: If `Iota` fails? Do we just get a complaint from `git`?

scripts/bumper/main.go
```diff
@@ -375,10 +382,43 @@ func detectNewCommits(ctx context.Context, logger *logrus.Entry, stagingDir, cen
...
+		if len(indices) == 0 {
+			break
+		}
+	}
+
+	// our ordered list is descending, but we need to cherry-pick from the oldest first
+	var reversedCommits []commit
+	for i := range orderedCommits {
+		reversedCommits = append(reversedCommits, orderedCommits[len(orderedCommits)-i-1])
+	}
```
tmshort, line 420: No easier way to reverse a list (array)?

tmshort:
A few minor comments. But it seems you are trying to pull in the same improvement I did in #578 ?

scripts/bumper/main.go
stevekuznetsov:
```
// Itoa is equivalent to FormatInt(int64(i), 10).
func Itoa(i int) string {}
```

What failure specifically are you worried about?

stevekuznetsov:
@tmshort yes, as you can see in my original bump PR I ran into it then :)

scripts/bumper/main.go
stevekuznetsov: What is difficult about this, or, how would you want to improve it?

scripts/bumper/main.go
tmshort: NVM... I was thinking it was `Atoi`...

scripts/bumper/main.go
tmshort:
```
slices.Reverse(orderedCommits)
```
Might need to use `golang.oef/x/exp/slices` in 1.20; `slices` is part of the standard library in 1.21

tmshort: Or even do it in place rather than doing an `append`. Unless you need to keep `orderedCommits` around unchanged?

stevekuznetsov: Eh, it's a short-lived CLI tool. The extra memory doesn't really matter. Happy to use the slices API when it lands in the stdlib.

ncdc:
Manually merging as e2e failures are unrelated to this PR

PR #581, opened 2023-10-11
Skuznets/bump vendor

stevekuznetsov:

PR #594, opened 2023-10-18
Downstream sync

tmshort:


dtfranz:
I'll run this once more but looks like we may have some investigating to do
/test e2e-upgrade

tmshort:
> I'll run this once more but looks like we may have some investigating to do

This is basically @stevekuznetsov's #581, and I think e2e-upgrade may be broken in by these PRs.

tmshort:
Closing due to #581 merging.

PR #595, opened 2023-10-20
OCPBUGS-22217: Fixing flake tests so they don't fail 100% of the time.

tmshort:
E2E tests were updated recently to use a generated namespace with an OperatorGroup created from the beginning of each test. However some of the tests were not updated to reflect this and were still creating and deleting resources in an additional namespace, or began to fail because the test environment required that no OperatorGroups exist from the beginning of the test.

This commit addresses two tests that have been failing at a rate of 100% due to the issues stated above.

Upstream-commit: 91b5762a9


tmshort:
"Flake" passed!

PR #598, opened 2023-10-25
try to constrain rbac based on e2e usage

awgreene:


manifests/0000_50_olm_01-olm-operator.serviceaccount.yaml
```diff
@@ -15,11 +15,106 @@ metadata:
...
+    resources:
+    - kialis
+    verbs:
+    - "*"
+  - apiGroups:
+    - monitoring.kiali.io
+    resources:
+    - monitoringdashboards
+    verbs:
+    - "*"
```
awgreene, line 117: The RBAC proposed in this PR stemmed from a run of audit2RBAC against an e2e test's audit logs, I'm not sure how we use these RBACs.
```
- apiGroups:
    - example.com
    resources:
    - tests
    verbs:
    - "*"
  - apiGroups:
    - kiali.io
    resources:
    - kialis
    verbs:
    - "*"
  - apiGroups:
    - monitoring.kiali.io
    resources:
    - monitoringdashboards
    verbs:
    - "*"
   ```

dtfranz: How confident can we be here that this audit will have captured all of the permissions we would need? I'm seeing kiali operator stuff in there and wondering if that means we would need to add every resource type from every operator that OLM manages in order to maintain functionality, which would be impractical.

dinhxuanvu: I’m not sure what other operators APIs has to do with this PR. This PR is to restrict the RBAC that is currently used by olm-operator to avoid using wildcards which means it can access a lot of other APIs that it doesn’t need. This only restricts it to a set of APIs that it needs.

dinhxuanvu: I don’t know why those kialia APIs were added in the original version of this PR but they are not needed as far as I know unless all the sudden OLM is using Kiali operator now.

dinhxuanvu: Also this PR should be on upstream as well. We can certainly refine/restrict RBAC on manifests that are used on upstream.

awgreene: Hey! So I saw a similar issue upstream and it's because of the CR/CRD validation logic that OLM has. As @dinhxuanvu pointed out, we had [an upstream PR ](https://github.com/operator-framework/operator-lifecycle-manager/pull/3076)that was a bit less strict on the RBAC we requested, I primarily was using this PR to test something on an OpenShift cluster.

The upstream PR linked earlier is being downstreamed in #600, so I'm going to close this PR.

awgreene:
I was using this PR to test some changes to OLM's RBAC, which lead to [this upstream PR](https://github.com/operator-framework/operator-lifecycle-manager/pull/3076) which is being downstreamed in #600, so I'm going to close this PR.

PR #604, opened 2023-11-08
OPRUN-3063: Generate OLM microshift manifests

jmprusi:
This PR adds the necessary changes to 'scripts/generate_crds_manifests.sh' to create a new set of manifests specifically targeted for microshift.

### DIFF
--- a/microshift-manifests/0000_50_olm_00-pprof-config.yaml
+++ b/microshift-manifests/0000_50_olm_00-pprof-config.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    release.openshift.io/create-only: "true"
+    capability.openshift.io/name: "OperatorLifecycleManager"
+  name: collect-profiles-config
+  namespace: openshift-operator-lifecycle-manager
+data:
+  pprof-config.yaml: |
+    disabled: False
--- a/scripts/generate_crds_manifests.sh
+++ b/scripts/generate_crds_manifests.sh
@@ -11,7 +11,6 @@ YQ="go run ./vendor/github.com/mikefarah/yq/v3/"
 CONTROLLER_GEN="go run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen"
 HELM="go run helm.sh/helm/v3/cmd/helm"
 
-
 ver=${OLM_VERSION:-"0.0.0-dev"}
 tmpdir="$(mktemp -p . -d 2>/dev/null || mktemp -d ./tmpdir.XXXXXXX)"
 chartdir="${tmpdir}/chart"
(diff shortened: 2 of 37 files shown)


ncdc:
@jmprusi please split this into 2 commits - 1 for the script edits, 1 for the generated files. Thanks!

scripts/generate_crds_manifests.sh
```diff
@@ -497,3 +494,52 @@ find "${ROOT_DIR}/manifests" -type f -exec $SED -i "1{/---/d}" {} \;
...
+# Create the kustomization file
+cat << EOF > "${ROOT_DIR}/microshift-manifests/kustomization.yaml"
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+EOF
+
+# Now let's generate the kustomization.yaml file for microshift to pick up the manifests.
+microshift_manifests_files=$(find "${ROOT_DIR}/microshift-manifests" -type f -name "*.yaml")
+# filest to ignore, substring match.
```
ncdc, line 524:
```suggestion
# files to ignore, substring match.
```

jmprusi:
@ncdc done, PTAL.

ncdc:
Need to fix verify failure

jmprusi:
> Need to fix verify failure

`make verify` returns 0 locally. looks like we are having some CI issues...

ncdc:
It's complaining about kustomization.yaml

jmprusi:
> It's complaining about kustomization.yaml

Fixed... It looks like `find` ordering is different in Macos vs Linux.

scripts/generate_crds_manifests.sh
```diff
@@ -495,3 +492,54 @@ find "${ROOT_DIR}/manifests" -type f -exec $SED -i "1{/---/d}" {} \;
...
 ${YQ} delete --inplace -d'*' manifests/0000_50_olm_00-namespace.yaml 'metadata.labels."pod-security.kubernetes.io/audit*"'
+
+
+# Let's copy all the manifests to a separate directory for microshift
+mkdir -p "${ROOT_DIR}/microshift-manifests/"
+rm -rf "${ROOT_DIR}/microshift-manifests/"*
+cp "${ROOT_DIR}"/manifests/* "${ROOT_DIR}/microshift-manifests/"
+
+# Let's add the annotations for ibm-cloud-managed on the standard manifests
+# !! Order matters, as we don't need the IBM managed annotations on the microshift manifests.
```
ncdc, line 503: What would happen if we did have these annotations on the manifests used by MicroShift? I'm thinking the less we differ, the better?

microshift-manifests/0000_50_olm_00-pprof-config.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    release.openshift.io/create-only: "true"
```
ncdc, line 7: This is an example of 1 file where the regular manifest has `capability.openshift.io/name: "OperatorLifecycleManager"` but the MicroShift variant does not. I think this capability annotation should be on the MicroShift manifests too?

scripts/generate_crds_manifests.sh
jmprusi: done, PTAL

microshift-manifests/0000_50_olm_00-pprof-config.yaml
jmprusi: done, PTAL

PR #614, opened 2023-11-20
OCPBUGS-23508: [release-4.14] Use generated namespaces in e2e tests

awgreene:
This commit updates all of OLM e2e test suites to generate a namespace rather than using a hard coded test namespace passed into the test as a parameter. The motivation for this change include:
1. Decreasing the opportunity for tests to influence each other. Tests are not guaranteed to cleanup after themselves and may influence the results of other tests.
2. Some distributions of OLM run alongside other controllers which may revert changes to resources in predetermined namespaces.

Upstream-commit: 108d6db0d11d5224214904c02e91391b0aada286

### DIFF
--- a/staging/operator-lifecycle-manager/test/e2e/catalog_e2e_test.go
+++ b/staging/operator-lifecycle-manager/test/e2e/catalog_e2e_test.go
@@ -47,9 +47,9 @@ const (
 
 var _ = Describe("Starting CatalogSource e2e tests", func() {
 	var (
-		ns  corev1.Namespace
-		c   operatorclient.ClientInterface
-		crc versioned.Interface
+		generatedNamespace corev1.Namespace
+		c                  operatorclient.ClientInterface
+		crc                versioned.Interface
 	)
 
 	BeforeEach(func() {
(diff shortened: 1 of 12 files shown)


awgreene:
Added the `NO-ISSUE:` prefix as this is just adding an e2e test.

staging/operator-lifecycle-manager/test/e2e/catalog_e2e_test.go
```diff
@@ -47,9 +47,9 @@ const (
 
 var _ = Describe("Starting CatalogSource e2e tests", func() {
 	var (
-		ns  corev1.Namespace
-		c   operatorclient.ClientInterface
-		crc versioned.Interface
+		generatedNamespace corev1.Namespace
```
everettraven, line 50: Is this name change to ensure commonality between all the tests and the name of the variable they all use for the namespace? It's definitely more descriptive but it seems like the anticipated functionality is already present in this file without the rename.

everettraven (approved):
One general comment, but it shouldn't block this from merging. Not sure if I have the proper permissions on this repo, but LGTM.

/lgtm

staging/operator-lifecycle-manager/test/e2e/catalog_e2e_test.go
awgreene: Hey, great callout! Your intuition is correct and this was discussed in the upstream PR. I should have introduced two commits:

- A commit that standardized the namespace variable name.
- A commit that introduced generated namespaces to e2e suites that didn't generate namespaces for the tests.

Ultimately it was decided to be non-blocking.

PR #626, opened 2023-11-30
Add ownership annotations to new and existing olm-managed secrets

dinhxuanvu:
As a part of certificates ownership work, all of new and existing secrets that are created with certificate information need to have ownership annotations. New secrets that are created with OLM certresources controller will have new ownership annotations injected at creation/update time. The existing olm-managed secrets that don't have the required annotations will get reconciled at startup when olm operator is restarted/redeployed. This PR only add OpenShift owning component annotation and the description annotation can be added later.


tmshort:
This contains non-upstreamed changes to the staging directory.

Because of that, it needs to be marked specially. I believe `[carry]` needs to be added to the git commit headline.

staging/operator-lifecycle-manager/pkg/controller/operators/olm/operator.go
```diff
@@ -2900,3 +2900,27 @@ func (a *Operator) ensureLabels(in *v1alpha1.ClusterServiceVersion, labelSets ..
...
+
+// syncSecret adds required ownership annotations to olm-managed secrets
+func (a *Operator) EnsureSecretOwnershipAnnotations() error {
+	secrets, err := a.lister.CoreV1().SecretLister().List(labels.SelectorFromSet(labels.Set{install.OLMManagedLabelKey: install.OLMManagedLabelValue}))
+	if err != nil {
+		return err
+	}
+	for _, secret := range secrets {
+		if secret.Annotations[install.OpenShiftComponent] == "" {
+			secret.Annotations[install.OpenShiftComponent] = install.OLMOwnershipAnnotation
```
joelanford, line 2912: I think this question applies here and in the cert generation code: Should we really say that the openshift component for every secret that OLM manages on behalf of operators is an OLM component?

Or could we properly associate the secrets we manage for the `Foo` operator to the `Foo` component?

dinhxuanvu: While I see your point, it doesn't change the fact that OLM is creating these secrets (and the certificates as well). Until OLM can delegate that responsibility to someone else, OLM effectively owns those resources from certificate ownership perspective. Also, the operators don't control these secrets. They are the consumers from that perspective. It is hardly logical for them to claim ownership on these resources that they don't have control over.

dinhxuanvu:
@joelanford @tmshort Seeking for a LGTM here unless there are any further feedbacks.

tmshort (approved):
/lgtm

PR #636, opened 2023-12-14
[release-4.15] OCPBUGS-25367: [CARRY] Fix panic issue when annotations map is nil

dinhxuanvu:
In some cases, secrets do not have any annotations and the annotations map are nil. The annotations map needs to be initialized to avoid panic error.


dinhxuanvu:
@perdasilva Seeking for backport-risk-assessed label. Thanks.

PR #652, opened 2024-01-09
Propagate grpc update

tmshort:
Upstream OLM removed the `replace` directive in go.mod for grpc. The same needs to happen here.
Run `go mod tidy/vendor/verify`

This should allow #651 to succeed.


tmshort:
Closing in favor of #653.

PR #690, opened 2024-02-12
Add relatedObjects to the initially created ClusterOperator

deads2k:
This allows for oc adm inspect and must-gather to collected debugging infomation in cases where the operator fails to run properly.


joelanford:
Seeing something changed in the `staging` directory tells me that these manifests originate upstream.

I'll make a similar PR there. Should we get that merged and let the fix come in via the normal sync process?

joelanford:
Thanks @deads2k! This is coming in via upstream sync in https://github.com/openshift/operator-framework-olm/pull/695

PR #698, opened 2024-02-20
TEMP: Test

tmshort:


tmshort:
It appears that the CSV queueinformer syncer is having a race condition with itself, which may have been made worse by adding a finalizer to a controller-runtime reconciler. Basically, it appears that `syncClusterServiceVersion()` takes a while to do its job, such that `handleClusterServiceVersionDeletion()` may occur at the same time; all while the controller-runtime reconciler with the finalizer has run. This problem manifests itself when manipulating RBAC. 

Via logging (as in this PR), I can see that modifications to the CSV occur, and then a delete occurs. The finalizer code cleans up the RBAC, but in the meantime, the `syncClusterServiceVersion()` function is still running, and tries to update the RBAC of the now deleted CSV using locally cached CSV data.

This causes the "Create CSV with permissions" test to fail with outstanding ClusterRoles.

The obvious fix is to consolidate the reconcilers into a controller-runtime (modern) reconciler, but that is _way_ too much effort and risk.

My solution is to check the DeletionTimestamp in the `syncClusterServiceVersion()` function multiple times, as it seems to take a while for that function to actually run (I will be adding logs to test this hypothesis). I might just need to get a fresh CSV, rather than rely on something from the queue...

dinhxuanvu:
> It appears that the CSV queueinformer syncer is having a race condition with itself, which may have been made worse by adding a finalizer to a controller-runtime reconciler. Basically, it appears that `syncClusterServiceVersion()` takes a while to do its job, such that `handleClusterServiceVersionDeletion()` may occur at the same time; all while the controller-runtime reconciler with the finalizer has run. This problem manifests itself when manipulating RBAC.
> 
> Via logging (as in this PR), I can see that modifications to the CSV occur, and then a delete occurs. The finalizer code cleans up the RBAC, but in the meantime, the `syncClusterServiceVersion()` function is still running, and tries to update the RBAC of the now deleted CSV using locally cached CSV data.
> 
> This causes the "Create CSV with permissions" test to fail with outstanding ClusterRoles.
> 
> The obvious fix is to consolidate the reconcilers into a controller-runtime (modern) reconciler, but that is _way_ too much effort and risk.
> 
> My solution is to check the DeletionTimestamp in the `syncClusterServiceVersion()` function multiple times, as it seems to take a while for that function to actually run (I will be adding logs to test this hypothesis). I might just need to get a fresh CSV, rather than rely on something from the queue...

This is not a new problem with OLM v0. Historically, OLM has multiple control loops that handle/touch the same resource type which leads so occasional race conditions such as the controller handling stale objects as they are being modified by another controller. I don't recall where this finalizer controller is being added (assuming on operator-controller repo) but you may want to strip off some of functionalities on RBAC handling from this olm-operator control-loop to the new place wherever that may be. On the other hands, you can avoid using cache on those two sync funcs and use direct requests to API server to avoid stale data. However, that may increase API request counts toward API server. If it ends up hammering API server side, you end up with a bug request to fix it.

dinhxuanvu:
> Understood it's not a new problem; there's inherit risks in moving all the reconciliation around, so trying to avoid that. When this condition occurs, there is an error that the CRB cannot be found (i.e. only the CR is found), and perhaps that would help this problem. I understand that hitting the API server could be an issue with this fix as-is.

Perhaps, what you can do here is to modify this area of code a bit to simply handle the NotFound error when RBAC objects are already deleted. Then, adjusting the e2e test case accordingly for that. As long as the intended behavior hasn't changed, it is not an issue to modify the test. They ain't meant to be set in stone.

tmshort:
The following log shows that the sync CSV starts (1), and then the finalizer - on the other reconciler -  is run (2, 3) while the sync CSV is still running, which completes at (6). Line (5) is where I update from the `lister` (API call?), and it recognizes that the CSV is being deleted (so RBAC is not being updated). A next sync at (7) recognizes that the CSV is being deleted (8), so the sync terminates early (9).
```
1. time="2024-02-21T21:18:14Z" level=info msg="start syncing CSV" csv=nginxsk2cl-stable id=cPHtB namespace=install-plan-e2e-srkkc phase=Installing
2. {"level":"info","ts":"2024-02-21T21:18:14Z","logger":"controllers.operatorcondition-generator","msg":"running finalizer","name":"nginxsk2cl-stable","namespace":"install-plan-e2e-srkkc"}
3. {"level":"info","ts":"2024-02-21T21:18:14Z","logger":"controllers.operatorcondition-generator","msg":"Cleaning up resources after CSV deletion","name":"nginxsk2cl-stable","namespace":"install-plan-e2e-srkkc","selector":"olm.owner=nginxsk2cl-stable,olm.owner.kind=ClusterServiceVersion,olm.owner.namespace=install-plan-e2e-srkkc"}
4. time="2024-02-21T21:18:14Z" level=info msg="install strategy successful" csv=nginxsk2cl-stable id=1zeFP namespace=install-plan-e2e-srkkc phase=Installing strategy=deployment
5. time="2024-02-21T21:18:14Z" level=info msg="CSV is being deleted - not processing RBAC" csv=nginxsk2cl-stable opgroup=install-plan-e2e-srkkc-operatorgroup
6. time="2024-02-21T21:18:14Z" level=info msg="end syncing CSV" csv=nginxsk2cl-stable id=cPHtB namespace=install-plan-e2e-srkkc phase=Installing
7. time="2024-02-21T21:18:14Z" level=info msg="start syncing CSV" csv=nginxsk2cl-stable id=/6svt namespace=install-plan-e2e-srkkc phase=Installing
8. time="2024-02-21T21:18:14Z" level=info msg="CSV is being deleted" csv=nginxsk2cl-stable id=/6svt namespace=install-plan-e2e-srkkc phase=Installing
9. time="2024-02-21T21:18:14Z" level=info msg="end syncing CSV" csv=nginxsk2cl-stable id=/6svt namespace=install-plan-e2e-srkkc phase=Installing
```

dinhxuanvu:
> The test should be fine. Q: If I use the `lister`, does that cause an API server call, or just a refresh from the the cache?

`lister` is using cache. You will need to use client directly to hit API server.

tmshort:
> `lister` is using cache. You will need to use client directly to hit API server.

That's what I thought, and the cache seems to be updated. So, my solution is feasible.

PR #703, opened 2024-02-28
AUTH-482: set required-scc for openshift workloads

liouk:


liouk:
@perdasilva could you please advise on why `ci/prow/verify` might be failing? Those verifications all succeed locally; on CI there seems to be an issue with `0000_50_olm_00-clusterserviceversions.crd.yaml`, but I can't figure out what.

perdasilva:
> @perdasilva could you please advise on why `ci/prow/verify` might be failing? Those verifications all succeed locally; on CI there seems to be an issue with `0000_50_olm_00-clusterserviceversions.crd.yaml`, but I can't figure out what.

@liouk I gave you bad advice before. What you need to run is `make generate-manifests`. Then amend your commit with the changes from that call. Then run `make verify` just to :eyes: and it should pass in CI. Reach out to me on Slack if you're still running into issues.

perdasilva:
@liouk so, even running `make generate-manifests` didn't work. I've been investigating a bit and think it might be to do with a go version mismatch. Can you switch to 1.21 and re-rerun generate-manifests? Thank ought to fix the problem, but I'm still testing it in CI. I'll come back with confirmation.

perdasilva:
Seems like something is funky here - but I'm really not sure why. I'll check out the PR and try to repro locally

perdasilva:
looking like flakes - going to retest but also see if I can determine the source of the issue

PR #704, opened 2024-02-28
OCPBUGS-25019: building rhel8/rhel9/static clis in payload image

grokspawn:
This PR generates a new operator-framework-clis image which contains rhel8 & rhel9 FIPS-compliant binaries and an updated static non-FIPS-compliant binary:
```sh
-rwxr-xr-x. 1 root root  64M Mar  2 03:54 configmap-server
-rwxr-xr-x. 1 root root  63M Mar  2 03:50 configmap-server-rhel8
-rwxr-xr-x. 1 root root  79M Mar  2 03:55 darwin-amd64-opm
-rwxr-xr-x. 1 root root  41M Mar  2 03:54 initializer
-rwxr-xr-x. 1 root root  41M Mar  2 03:50 initializer-rhel8
-rwxr-xr-x. 1 root root  77M Mar  2 03:54 opm
-rwxr-xr-x. 1 root root  77M Mar  2 03:50 opm-rhel8
-rwxr-xr-x. 1 root root  75M Mar  2 03:57 opm-static
-rwxr-xr-x. 1 root root  41M Mar  2 03:54 registry-server
-rwxr-xr-x. 1 root root  41M Mar  2 03:50 registry-server-rhel8
-rwxr-xr-x. 1 root root 134M Mar  2 03:56 windows-amd64-opm
```
rhel8 binaries are provided only where the binary is dynamically linked for rhel9 in main/latest OCP, so for e.g. no darwin, windows versions.

Followups in other repos will lead to ART packaging binaries from this image to the mirrors site instead of the image used as the basis for catalog pods. 
(and possibly to slimming down the catalog basis image)

### DIFF
--- a/operator-framework-cli.Dockerfile
+++ b/operator-framework-cli.Dockerfile
@@ -1,0 +1,55 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.21-openshift-4.16 AS builder-rhel8
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
+WORKDIR /src
+COPY . .
+RUN make build/registry cross
+
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS builder
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
+WORKDIR /src
+COPY . .
+RUN make build/registry cross
+
+# copy and build vendored grpc_health_probe
+RUN CGO_ENABLED=0 go build -mod=vendor -tags netgo -ldflags "-w" ./vendor/github.com/grpc-ecosystem/grpc-health-probe/...
+
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS staticbuilder
+# Permit opm binary to be compiled statically. The Red Hat compiler
+# provided by ART will otherwise force FIPS compliant dynamic compilation.
+ENV GO_COMPLIANCE_EXCLUDE="build.*operator-registry/cmd/opm"
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
+WORKDIR /src
+COPY . .
+RUN CGO_ENABLED=0 make build/registry
+FROM registry.ci.openshift.org/ocp/4.16:base-rhel9
+
+COPY --from=builder /src/bin/* /bin/registry/
+COPY --from=builder /src/grpc-health-probe /bin/grpc_health_probe
+# copy the dynamically-linked versions to /bin/registry with a -rhel8 suffix
+COPY --from=builder-rhel8 /src/bin/opm /bin/registry/opm-rhel8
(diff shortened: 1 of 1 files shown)


grokspawn:
currently builds everything, but transplants only the linux opm binary from the rhel8 builder.
When I built the image, I do see both binaries:
```sh
-rwxr-xr-x. 1 root root  80551544 Feb 28 21:11 opm
-rwxr-xr-x. 1 root root  80310624 Feb 28 21:07 opm-rhel8
```
but the ldd output is not what I expected. I thought that the resulting container would be in the context of rhel9 so the rhel8 binary would have some dangling library links, but they all appear to be satisfied:
```sh
bash-5.1$ ldd opm 
	linux-vdso.so.1 (0x00007ffe1c9a2000)
	libresolv.so.2 => /lib64/libresolv.so.2 (0x00007f8dea739000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f8dea530000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f8dea752000)
bash-5.1$ ldd opm-rhel8 
	linux-vdso.so.1 (0x00007ffe9ecb3000)
	libresolv.so.2 => /lib64/libresolv.so.2 (0x00007fbc56463000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fbc5645e000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007fbc56459000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fbc56250000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fbc5647c000)
```

These results look strange because both sets are resolved (with different hex values for their requirements) but also because there are more linkages in the rhel8 version.  

It's probably a product of the platform version on which they were built, but hopefully we can demystify the issues.

grokspawn:
/retest 
PR has no code.

grokspawn:
/retest 
infra problems in openshift/test/origin

grokspawn:
Failing e2e is waiting for infra PR to merge: https://github.com/openshift/origin/pull/28631

tmshort:
> It's probably a product of the platform version on which they were built, but hopefully we can demystify the issues.

Did RHEL9 combine `libdl` and `libpthread` into `libc`?

grokspawn:
> Did RHEL9 combine libdl and libpthread into libc?
Great question. 

It looks like at least libdl was:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/considerations_in_adopting_rhel_9/index
> The libdl library now merged into libc.so.6
>
>In RHEL 8, the libdl library was a distinct library. In RHEL 9, the libdl library has been merged into the core C library libc.so.6. This means that interposing the dlsym function is now much more difficult. Applications that need to control the way how symbol resolution works should switch to the auditor (LD_AUDIT) interfaces.

joelanford:
If I'm understanding correctly, this will increase the size of all of the catalog images by the same amount as the binary size, which impacts mirroring and image pulling.

Is it possible to build a separate image for the RHEL8 binary?

grokspawn:
/hold
We're having conversations which can change direction of this effort

grokspawn:
/hold cancel
switched approaches to a new "CLIs" image.

everettraven (approved):
/lgtm

PR #705, opened 2024-02-28
(psm) Wait for required RBAC before creating packageserver CSV

anik120:
Motivation: https://issues.redhat.com/browse/OCPBUGS-23744 PackerServer ClusterOperator becomes unavailable momentarily on OCP upgrades, due to initial authentication blips. Even though the ClusterOperator heals eventually, ClusterOperator going down requires cluster admins to [react immediately](https://github.com/openshift/api/blob/c3f7566f6ef636bb7cf9549bf47112844285989e/config/v1/types_cluster_operator.go#L149-L153). Admins should not be paged for something that we know will heal eventually, This also shows up ~30% of the time in OCP CI as a failure since the healing does not take place within the allocated wait time.

### DIFF
--- a/pkg/package-server-manager/controller.go
+++ b/pkg/package-server-manager/controller.go
@@ -27,6 +27,8 @@ import (
 	"github.com/openshift/operator-framework-olm/pkg/manifests"
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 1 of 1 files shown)


pkg/package-server-manager/controller.go
```diff
@@ -102,6 +108,21 @@ func (r *PackageServerCSVReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	log.Info("checking to see if required RBAC exists")
+	if err := client.Get(ctx, types.NamespacedName{Name: "olm-operator-serviceaccount", Namespace: namespace}, &corev1.ServiceAccount{}); err != nil {
+		return fmt.Errorf("could not get service account:%v", err)
+	}
+	if err := client.Get(ctx, types.NamespacedName{Name: "system:controller:operator-lifecycle-manager"}, &rbacv1.ClusterRole{}); err != nil {
+		return fmt.Errorf("could not get ClusterRole:% v", err)
+	}
+	if err := client.Get(ctx, types.NamespacedName{Name: "olm-operator-binding-openshift-operator-lifecycle-manager"}, &rbacv1.ClusterRoleBinding{}); err != nil {
+		return fmt.Errorf("could not get ClusterRoleBinding: %v", err)
+	}
```
joelanford, line 121: Does our cache already have ServiceAccounts, ClusterRoels and ClusterRoleBindings in it?

If not, this will likely cause our controller to cache all of them.

I'd suggest either configuring the cache to select _just_ on these particular names with a field selector or to use the direct APIReader client.

joelanford: See here: https://github.com/openshift/operator-framework-olm/blob/b3bfaa4b69bf7dfe8254cdadd8a6c8fc450e2b97/cmd/package-server-manager/main.go#L95-L104

PR #708, opened 2024-03-01
OCPBUGS-23744: Wait for required RBAC before creating packageserver CSV

kevinrizza:
Motivation: https://issues.redhat.com/browse/OCPBUGS-23744 PackerServer ClusterOperator becomes unavailable momentarily on OCP upgrades, due to initial authentication blips. Even though the ClusterOperator heals eventually, ClusterOperator going down requires cluster admins to [react immediately](https://github.com/openshift/api/blob/c3f7566f6ef636bb7cf9549bf47112844285989e/config/v1/types_cluster_operator.go#L149-L153). Admins should not be paged for something that we know will heal eventually, This also shows up ~30% of the time in OCP CI as a failure since the healing does not take place within the allocated wait time.

Supersedes #705

### DIFF
--- a/cmd/package-server-manager/main.go
+++ b/cmd/package-server-manager/main.go
@@ -11,6 +11,9 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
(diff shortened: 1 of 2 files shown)


anik120:
@joelanford fyi

PR #713, opened 2024-03-07
image just as a cli source for mirror publishing

grokspawn:
strips out runnable aspects and list of CLIs so this is just a source of OLM/OPM clients for now.


operator-framework-cli.Dockerfile
```diff
@@ -12,44 +12,18 @@ WORKDIR /src
...
 
 USER 1001
-EXPOSE 50051
-
-ENTRYPOINT ["/bin/registry-server"]
-CMD ["--database", "/bundles.db"]
 
-LABEL io.k8s.display-name="OpenShift Operator Registry" \
-      io.k8s.description="This is a component of OpenShift Operator Lifecycle Manager and is the base for operator catalog API containers." \
+LABEL io.k8s.display-name="OpenShift Operator Lifecycle Manager CLIs" \
```
joelanford, line 26:
```suggestion
LABEL io.k8s.display-name="OpenShift Operator Framework CLIs" \
```

operator-framework-cli.Dockerfile
```diff
@@ -12,44 +12,18 @@ WORKDIR /src
...
+      io.k8s.description="This is a component of OpenShift Operator Lifecycle Manager and is used to manipulate related artifacts, including catalogs of operators." \
```
joelanford, line 27:
```suggestion
      io.k8s.description="This is a non-runnable image containing binary builds of various Operator Framework CLI tools, primarily used to publish binaries to the OpenShift mirror." \
```

operator-framework-cli.Dockerfile
```diff
@@ -12,44 +12,18 @@ WORKDIR /src
...
       maintainer="Odin Team <aos-odin@redhat.com>" \
-      summary="Operator Registry runs in a Kubernetes or OpenShift cluster to provide operator catalog data to Operator Lifecycle Manager."
+      summary="Operator Lifecycle Manager CLIS are used to manipulate OLM infrastructure, including catalogs of operators."
```
joelanford, line 29: Do we know what the difference is between `io.k8s.description` and `summary`?

This suggestion I'm making would make those the same, but not sure if that's actually a good idea.
```suggestion
      summary="This is a non-runnable image containing binary builds of various Operator Framework CLI tools, primarily used to publish binaries to the OpenShift mirror."
```

operator-framework-cli.Dockerfile
joelanford, line 1: Pluralize to `operator-framework-clis.Dockerfile`?

operator-framework-cli.Dockerfile
grokspawn: Not enough samples to have an opinion here. 
oc clients doesn't use `summary` (https://github.com/openshift/oc/blob/master/images/cli-artifacts/Dockerfile.rhel#L31-L33), so on the off-chance that this turns out to be a short-lived image, let's simplify.

joelanford (approved):

PR #714, opened 2024-03-11
NO-ISSUE: Semver lastedge error (#1169)  [sync]

grokspawn:
* when the last edge in a bundle set is +Y the existing code does not clear the skips list and crosses Y thresholds counter to design intent missing proper last-max-z detection for replaces
working at long last

* re-add cumulative skips over y changes in an x stream so folks can always skip directly to channel head

---------

Upstream-commit: 2096c4f5d0d2cb3ed31767d406afe10786565f01


tmshort (approved):
/lgtm

PR #719, opened 2024-03-27
OCPBUGS-31080,OCPBUGS-31438: Manual downstream sync

dtfranz:
Manually syncing since a reverted dependabot commit is breaking the syncer tool at the moment.


grokspawn:
@dtfranz `make verify` works on your local?

PR #724, opened 2024-03-29
OCPBUGS-31547: UPSTREAM: <carry>: disable collect-profiles cronjob

kevinrizza:
We've made some significant perf improvements in the last release and this job is challenging to maintain. Disabling it for now, but leaving the bits to potentially serve it if we want to manually configure the job on a customer cluster

This has historically been difficult to explain to customers and at the moment provides tenuous value


jianzhangbjz:
Test pass, details: https://issues.redhat.com/browse/OCPBUGS-31547
/lgm
/label qe-approved

kevinrizza:
/retest

these failures seem like cluster initialization issues, not something specific to this pull request

tmshort:
We're just disabling? Not removing the code as well?

kevinrizza:
> We're just disabling? Not removing the code as well?

The intention is to prevent the cronjob from running so that this is no longer always happening on every cluster. It still seems useful to me as a potential data point if we want to do profiling on a specific customer cluster without having to give them special compiled bits

perdasilva:
Failure looks legit:

```
{  fail [github.com/openshift/origin/test/extended/operators/certs.go:115]: Unexpected error:
    <errors.aggregate | len:1, cap:1>: 
    secrets/pprof-cert[openshift-operator-lifecycle-manager] MISSING tls.crt content
    
    [
        <*errors.errorString | 0xc001a936e0>{
            s: "secrets/pprof-cert[openshift-operator-lifecycle-manager] MISSING tls.crt content\n",
        },
    ]
occurred
Ginkgo exit error 1: exit with code 1}
```

PR #730, opened 2024-04-10
[WIP] OCPBUGS-31547: UPSTREAM: <carry>: disable collect-profiles cronjob

perdasilva:
We've made some significant perf improvements in the last release and this job is challenging to maintain. Disabling it for now, but leaving the bits to potentially serve it if we want to manually configure the job on a customer cluster.

For the microshift manifests, deleting the manifest entirely because there is no cvo for microshift


perdasilva:
This is a clone + (possible) fix of #724

tmshort:
The tests are failing pretty early, all with the same reason.

PR #748, opened 2024-05-20
Fix generate CRD script by using copy

tmshort:
Using a symlink to point inside the staging directories seems to confuse newer versions of sigs.k8s.io/controller-tools.

Specifically, the set up of a temporary directory for controller-gen uses a symlink into staging/api, which confuses _something_ in golang to use ${ROOT}/staging/api/go.mod rather than ${ROOT}/go.mod, causing it to return errors that indicate go.mod inconsistencies with vendor/modules.txt. This would normally be fixable via the `go mod` commands, but looking at the wrong go.mod is not a normal situation.


operator-framework-tools.Dockerfile
```diff
@@ -1,11 +1,16 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.21-openshift-4.16 AS builder-rhel8
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.21-openshift-4.17 AS builder-rhel8
 ENV GOPATH /go
 ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
 WORKDIR /src
 COPY . .
+# Install golang 1.22.2
```
grokspawn, line 6: /hold
ART is generating a new `registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.17` we can use here to coordinate with new go1.22 toolchain.

grokspawn, line 6: ART is generating a new registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.22-openshift-4.17 we can use here to coordinate with new go1.22 toolchain.

grokspawn (changes requested):

PR #777, opened 2024-06-11
OCPBUGS-35305: [release-4.15] catalog-operator: delete catalog pods stuck in Terminating state due to unreachable node

joelanford:
Upstream-commit: 68c24cf94371f7a73619966f9e49612fd16fa76a


perdasilva:
was trying to fix this but it seems it needs go1.21

PR #802, opened 2024-06-27
OCPBUGS-35970: switch base image

grokspawn:


grokspawn:
/hold Joe requested an entrypoint to try to be clear we intend this image not to be executable
IMO the best way to do this is to base on scratch, but evidently that is not allowed.

grokspawn:
/hold cancel

got the entrypoint/cmd clearers there, plus mass whitespacing of unrelated content

grokspawn:
/retest
some infra problems this morning, unrelated to the PR

PR #807, opened 2024-07-01
OCPBUGS-30333: Remove Helm

bentito:
Replaces all the manifests with required intermediate state versions. They then get further munging via yq and script.


bentito:
Closing in favor of #837

PR #837, opened 2024-08-09
OCPBUGS-30333: Move Helm to bingo dep and update it

bentito:
We're adding `bingo` to the top level of this monorepo. In doing this we are also changing the dependency to a version of Helm that is free of the target bug.

Actual files in need of review:
.bingo/*
Makefile
go.mod
~go.sum~
scripts/generate_crds_manifests.sh
tools.go


bentito:
@joelanford I think this is due to the toolchain version and go version maybe not being in sync? from the `verify` failed test:

`./scripts/generate_crds_manifests.sh: line 69: /go/bin/helm-v3.15.2: No such file or directory`

grokspawn:
lgtm, but I'll just stick one label on here in case someone else wants to 👀 today.  Otherwise I'll try to circle back around and merge. 

/approve

PR #838, opened 2024-08-09
OCPBUGS-38290: [release-4.16] (fix) Resolver: list CatSrc using client, instead of referring to registry-server cache

anik120:

                "status": "True",
                "type": "ResolutionFailed"


tmshort (approved):
/lgtm

perdasilva:
@anik120 you may need to edit the PR title removing the space preceding the ':'

anik120:
@perdasilva thanks for catching that! Ticket looks like is linked correctly now, just have to wait for the 4.17 bug to be verified.

PR #842, opened 2024-08-15
[release-4.14] OCPBUGS-38544: (fix) Resolver: list CatSrc using client, instead of referring to registry-server cache (#3349)

anik120:

                "status": "True",
                "type": "ResolutionFailed"


tmshort:
The failing tests have:
```
  * could not initialize namespace: timed out waiting for image pull secrets 
```

PR #843, opened 2024-08-16
[release-4.13] OCPBUGS-38608: (fix) Resolver: list CatSrc using client, instead of referring to registry-server cache (#3349)

anik120:

                "status": "True",
                "type": "ResolutionFailed"

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/registry/resolver/step_resolver_test.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/registry/resolver/step_resolver_test.go
@@ -335,6 +335,7 @@ func TestResolver(t *testing.T) {
 			name: "SubscriptionOmitsChannel",
 			clusterState: []runtime.Object{
 				newSub(namespace, "package", "", catalog),
+				newOperatorGroup("foo", namespace),
 			},
 			bundlesByCatalog: map[resolvercache.SourceKey][]*api.Bundle{
 				catalog: {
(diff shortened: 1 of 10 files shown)


staging/operator-lifecycle-manager/pkg/controller/registry/resolver/step_resolver_test.go
```diff
@@ -482,6 +489,7 @@ func TestResolver(t *testing.T) {
 			name: "SingleNewSubscription/ResolveOne/BundlePath",
 			clusterState: []runtime.Object{
 				newSub(namespace, "a", "alpha", catalog),
+				newOperatorGroup("foo", namespace),
```
anik120, line 492: Had to had a few of these extra from the original commit. Looks like they were added in a different commit from Ankita but that was back ported till 4.14, so 4.13 was missing them and this commit needed them.

staging/operator-lifecycle-manager/pkg/controller/registry/resolver/step_resolver_test.go
```diff
@@ -528,6 +536,7 @@ func TestResolver(t *testing.T) {
 			name: "SingleNewSubscription/ResolveOne/AdditionalBundleObjects",
 			clusterState: []runtime.Object{
 				newSub(namespace, "a", "alpha", catalog),
+				newOperatorGroup("foo", namespace),
```
anik120, line 539: eg

staging/operator-lifecycle-manager/pkg/controller/registry/resolver/step_resolver_test.go
```diff
@@ -572,6 +582,7 @@ func TestResolver(t *testing.T) {
 			name: "SingleNewSubscription/DependencyMissing",
 			clusterState: []runtime.Object{
 				newSub(namespace, "a", "alpha", catalog),
+				newOperatorGroup("foo", namespace),
```
anik120, line 585: and this...etc..

anik120:
@Xia-Zhao-rh can we get the `cherry-pick-approved` label on this please

PR #846, opened 2024-08-30
Add bingo-update target

tmshort:
This is required for generate-manifests that use a specific version of helm

Because this might be run in a container that doesn't already have bingo (i.e. operator-framework-tooling, a.k.a. bumper), install bingo as well.


tmshort:
This needs to be done as part of the bumper, AFAICT

PR #870, opened 2024-09-17
OCPBUGS-42114: add optional schema migrations; default to olm.bundle.object instead of olm.csv.metadata

grokspawn:
… (#1384)

* Allow disabling of migrations for bundles and sqlite DBs
* intro new migrations package
* schema migration optional; default to olm.bundle.object
* review resolutions
* use a type to protect migration interactions
* review revisions
---------

Upstream-commit: c80e8751e716c069017b55279a0f29d0254d426a

### DIFF
--- a/staging/operator-registry/alpha/action/migrations/migrations.go
+++ b/staging/operator-registry/alpha/action/migrations/migrations.go
@@ -1,0 +1,109 @@
+package migrations
+
+import (
+	"fmt"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+)
+
+type MigrationToken string
+
+const (
+	invalidMigration string = ""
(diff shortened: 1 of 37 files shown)


staging/operator-registry/alpha/action/migrations/migrations.go
```diff
@@ -0,0 +1,109 @@
...
+
+type Migrations struct {
+	Migrations []Migration
+}
+
+// CloneAllMigrations returns a shallow copy of allMigrations
+// since slices.Clone is not available in the current golang version
+func CloneAllMigrations() []Migration {
+	return append(allMigrations[:0:0], allMigrations...)
+}
```
grokspawn, line 55: This is a semi-kludge to replace `slices.Clone()` from later golang versions. Hopefully not too odious.

PR #894, opened 2024-11-12
use ubi instead of scratch

jianzhangbjz:
I'm not sure if I should modify the downstream directly, but some users have conerns when using the `scratch`, especially when building the multi-arch image. They have to modify the Dockfile after `opm alpha bundle generate xxx`. They will get an error if they forget to update the base image. So, I just submitted this PR, any comments are welcome! Thanks!
```console
[yapei@yapei-mac apicast (operators)]$ opm alpha bundle validate --tag quay.io/yapei/apicast-operators-bundle:latest --image-builder podman
Command "validate" is deprecated, This subcommand is deprecated and will be removed in a future release. Migrate to operator-sdk bundle validate
INFO[0000] Create a temp directory at /var/folders/gw/6ckgw32n7rz4yncmkkpsvjl80000gn/T/bundle-2170207577  container-tool=podman
DEBU[0000] Pulling and unpacking container image         container-tool=podman
INFO[0000] running /opt/podman/bin/podman pull quay.io/yapei/apicast-operators-bundle:latest  container-tool=podman
ERRO[0002] Trying to pull quay.io/yapei/apicast-operators-bundle:latest...
Error: choosing an image from manifest list docker://quay.io/yapei/apicast-operators-bundle:latest: no image found in image index for architecture "amd64", variant "", OS "linux"  container-tool=podman
Error: error pulling image: Trying to pull quay.io/yapei/apicast-operators-bundle:latest...
Error: choosing an image from manifest list docker://quay.io/yapei/apicast-operators-bundle:latest: no image found in image index for architecture "amd64", variant "", OS "linux"
...
```

### DIFF
--- a/staging/operator-registry/pkg/lib/bundle/generate.go
+++ b/staging/operator-registry/pkg/lib/bundle/generate.go
@@ -335,7 +335,7 @@ func GenerateDockerfile(mediaType, manifests, metadata, copyManifestDir, copyMet
 	relativeMetadataDirectory = filepath.ToSlash(relativeMetadataDirectory)
 
 	// FROM
-	fileContent += "FROM scratch\n\n"
+	fileContent += "FROM registry.access.redhat.com/ubi9:latest\n\n"
 
 	// LABEL
 	fileContent += fmt.Sprintf("LABEL %s=%s\n", MediatypeLabel, mediaType)


grokspawn:
Hi @jianzhangbjz 
You're right that we don't want to modify ./staging contents, since they get sync'd from the CNCF project and could easily get stomped on. 
I'm of two minds on the subject of this alpha level feature needing support.  
I think it's OK to add a new, optional flag to provide the capability, still defaulting to the existing behavior, like [here](https://github.com/operator-framework/operator-registry/pull/1519). And that would be available in downstream `opm` versions at the latest unreleased OCP version.
But I think that we would not backport this feature functionality without a really compelling argument.

joelanford:
Bundles images explicitly should NOT be multi-arch images. They contain just plaintext files, so they are not platform-specific.

Only images that contain executable binaries need to be multi-arch. And the manifests inside of a bundle images should always reference the multi-arch manifest of the controller and related images.

Is the validate bundle command failing on scratch-based bundle images?

grokspawn:
Hi @jianzhangbjz.  I think we can close this PR.  We downstreamed [operator-framework/operator-registry/#1519](https://github.com/operator-framework/operator-registry/pull/1519) as #933 which would give folks the ability to substitute a different base image if they cannot use scratch.

jianzhangbjz:
Great! Thanks! I closed this one.

PR #898, opened 2024-11-19
OCPBUGS-44760: fix: call TokenRequest API when service account token secret is missing

grokspawn:
… (#3377)

This is a backport of the original commit. Original description follows:

* fix: call TokenRequest API when service account token secret is missing

Beyond Kubernetes 1.22, the service account token secret is not automatically, created. Therefore, when OLM is not able to find the service account token secret, it should request one from the k8s api server.

Ref: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#manual-secret-management-for-serviceaccounts

* fix: return error

* handle error when creating sa token from TokenRequest API fails

* move return from inner loop

Move the return to line 48. This will ensure a value is returned whether we successully create a service account token from the TokenRequest API or get an error

---------

Upstream-commit: e20778cbb590f82dbe9cba500e98e7317222e5b5


bentito (approved):
lgtm

PR #921, opened 2024-12-18
OCPBUGS-46595: CRD upgrade existing CR validation fix (#3442)

grokspawn:
_manual fix of failed automation from #917 going to release-4.14._

* fixed to pass map[string]interface instead of unstructured.Unstructured

We started seeing some issues with folks who had spurious CRD incompatibility claims when updating operators.  It is a failure in OLM code which validates existing CRs against incoming CRDs, recently updated in #3387.

This manifested in `InstallPlan` `.status.Message` something like:
```
retrying execution due to error: error validating existing CRs against new CRD's schema for \"pgadmins.postgres-operator.crunchydata.com\": error validating postgres-operator.crunchydata.com/v1beta1, Kind=PGAdmin \"openshift-operators/example-pgadmin\": updated validation is too restrictive: [].spec.tolerations[0].tolerationSeconds: Invalid value: \"number\": spec.tolerations[0].tolerationSeconds in body must be of type integer: \"number\"
```

The difference between the predecessor calling convention and the one introduced in #3387 appears to be that one is a pointer and the other is concrete.

old
```golang
unstructured.Unstructured{Object:map[string]interface...
```

new
```golang
&unstructured.Unstructured{Object:map[string]interface...
```

so it would seem that merely type-asserting the value and de-referencing it would yield the appropriate result, but it appears instead that it effectively disables all CR vs CRD reconciliation checks (evidenced by the fact that the unit tests multiply fail).

But k8s already dereferences pointer parameters [here](https://github.com/kubernetes/kube-openapi/blob/master/pkg/validation/validate/schema.go#L139-L141) during validation.  So that isn't it.

And the `validate.ValidateCustomResource` interface is terrifyingly permissive in allowing `customResource` as `interface{}` [here](https://pkg.go.dev/k8s.io/apiextensions-apiserver@v0.31.3/pkg/apiserver/validation#ValidateCustomResource). So we cannot derive guidance from it.

Taking a page from k8s' use of the validation API, which uses `unstructured.UnstructuredContent()` to convert the `unstructured.Unstructured` into a `map[string]interface{}` [here](https://github.com/kubernetes/kubernetes/blob/1504f10e7946f95a8b1da35e28e4c7453ff62775/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/validator.go#L54) then we achieve the desired results.

* adding tests

---------

Upstream-commit: 1cfabfe5a495fe3cb276fce93255cdfed7d60783

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/operators/catalog/operator.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/operators/catalog/operator.go
@@ -1918,15 +1918,15 @@ func validateExistingCRs(dynamicClient dynamic.Interface, gvr schema.GroupVersio
 		return dynamicClient.Resource(gvr).List(context.TODO(), opts)
 	}))
 	validationFn := func(obj runtime.Object) error {
+		// lister will only provide unstructured objects as runtime.Object, so this should never fail to convert
+		// if it does, it's a programming error
+		cr := obj.(*unstructured.Unstructured)
 		validator, _, err := validation.NewSchemaValidator(newCRD.Spec.Validation)
 		if err != nil {
 			return fmt.Errorf("error creating validator for schema %#v: %s", newCRD.Spec.Validation, err)
 		}
-		err = validation.ValidateCustomResource(field.NewPath(""), obj, validator).ToAggregate()
+		err = validation.ValidateCustomResource(field.NewPath(""), cr.UnstructuredContent(), validator).ToAggregate()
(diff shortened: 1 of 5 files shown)


Xia-Zhao-rh:
test PASS. detail https://issues.redhat.com/browse/OCPBUGS-46595
/label qe-approved
/label cherry-pick-approved
/lgtm

PR #929, opened 2025-01-09
[WIP] OCPBUGS-45845: Queue fix final

perdasilva:

PR #936, opened 2025-01-16
OCPBUGS-48486: Fix concurrent namespace resolution

perdasilva:


kuiwang02:
@perdasilva Hi, what is the different between this PR and #929 ?
Thanks

perdasilva:
> @perdasilva Hi, what is the different between this PR and #929 ? Thanks

@kuiwang02 there shouldn't be any differences. The reason I'm not using that one is that I just side-loaded the upstream PR (which hadn't merged yet) for your testing. After you gave your approval, we felt comfortable merging upstream, and this is us downstreaming it now =D I think it would be ok to do do a lite verification. Assuming CI goes through =S

tmshort:
Why isn't this an upstream fix?

tmshort:
Isn't this just the first commit of https://github.com/openshift/operator-framework-olm/pull/937 ? Now, why is that not sync'ing right? Perhaps we should do a manual sync and fix it up (vendor-wise, or whatever)?

tmshort:
@perdasilva the upstream PR https://github.com/operator-framework/operator-lifecycle-manager/pull/3483 has been merged.
PR #937 has upstream PR3483 as the first commit, but it is failing.
Why are we not able to use #937 as-is (via the bumper)? Why do we have to modify the vendor directory? 

I'm concerned this will bite us in the future.

@joelanford since this is a downstream of your PR, can you verify it looks correct?

perdasilva:
I mentioned it in the auto sync PR. We cannot bring it down because it requires downstream changes. The fix for the concurrent namespace resolution makes changes to the workqueues. We have some downstream only reconcilers. Merging this will clear up the auto sync.

tmshort (approved):
/lgtm
/approve

PR #960, opened 2025-02-14
OCPBUGS-32262: couldn''t find unpacked step

gallettilance:
# Description

Because of [changes](https://github.com/openshift/operator-framework-olm/commit/e248469f170ddb8ccf03fa714924e75efb9de946#diff-885d1e95e50eb8ece2643ba380c0449ab1adaeba14cf461a0a58b1afcc61f228so) to the hashing algorithm that identifies manifests that need to be applied during install / upgrade, the unpack steps for bundles that have been installed using the old hashing algorithm cannot be found as they have a different name.

This code reintroduces the legacy hashing algorithm to resolve only objects that have been installed using this legacy process so they are not stranded. It adds the legacy steps to the map of steps. If the install was made using the new algorithm, these olds steps are unused. If the install was made using the old algorithm, the new steps are unused.

### DIFF
--- a/staging/operator-lifecycle-manager/go.mod
+++ b/staging/operator-lifecycle-manager/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/containers/image/v5 v5.33.0
 	github.com/coreos/go-semver v0.3.1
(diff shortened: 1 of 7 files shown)


perdasilva:
Cloned to #974

PR #970, opened 2025-03-04
OCPBUGS-52320: [release-4.17] The openshift-operator-lifecycle-manager still use old pod-security.kubernetes.io/*-version v1.24 and v1.25 respectively

camilamacedo86:


camilamacedo86:
Closing since QE would like to set latest instead

PR #972, opened 2025-03-06
OCPBUGS-42526: Ensure that PSA label is latest instead of pinning versions

camilamacedo86:


jianzhangbjz:
Test pass, details: https://issues.redhat.com/browse/OCPBUGS-52340

anik120 (approved):
/approve 

Verified context from https://issues.redhat.com/browse/OCPBUGS-41849:

```
What
Set the version label of warn, audit and enforce to "latest" from "v1.24".

Why
We kept it at "v1.24" due to potential issues with HyperShift.
We are moving to "latest" to avoid the cost of maintaining it by-version.
It was discussed and decided in a meeting with David Eads (Wed Sep 12 2024).
```

camilamacedo86:
Hi @anik120 

> Verified context from https://issues.redhat.com/browse/OCPBUGS-41849:
> 
> What
> Set the version label of warn, audit and enforce to "latest" from "v1.24".
> 
> Why
> We kept it at "v1.24" due to potential issues with HyperShift.
> We are moving to "latest" to avoid the cost of maintaining it by-version.
> It was discussed and decided in a meeting with David Eads (Wed Sep 12 2024).


To clarify:
- We saw that, but we discussed internally and we first decided to ensure that we would pinned all version: See; https://redhat-internal.slack.com/archives/C06KP34REFJ/p1739887248046889?thread_ts=1739880491.760029&cid=C06KP34REFJ
- Then, after all we changed it because according to QE it will not be a problem. Because the latest hasn't been released yet,  we have control over the newest version in all cases. (latest available in the release)

PR #990, opened 2025-04-07
Bump olm for ConsolePlugin resource support

jpinsonneau:
Testing https://github.com/operator-framework/operator-lifecycle-manager/pull/3551 & https://github.com/operator-framework/operator-registry/pull/1634 changes using https://github.com/netobserv/network-observability-operator/pull/1374 bundle

Clusterbot
- build: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/1909607438126944256
- launch: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/1909607467831005184

Console plugin is created successfully however, it is disabled by default: 
image

Ideally that should be controlled by the Console Plugin Enabled toggle at operator installation:
image

When the operator is uninstalled, the ConsolePlugin resource is not removed despite the olm.managed label is present:
image

I wonder if `ownerReferences` would make sense here ?

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/operators/catalog/supportedresources.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/operators/catalog/supportedresources.go
@@ -10,6 +10,7 @@ const (
 	ConsoleQuickStartKind     = "ConsoleQuickStart"
 	ConsoleCLIDownloadKind    = "ConsoleCLIDownload"
 	ConsoleLinkKind           = "ConsoleLink"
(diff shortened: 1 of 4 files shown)


jpinsonneau:
Registry is [using go 1.23.7](https://github.com/operator-framework/operator-registry/blob/master/go.mod#L3) which breaks cluster bot build so I reverted that for testing

PR #1018, opened 2025-06-11
replace rhel8 builder pullspec after great purge

grokspawn:
CI failures around branch cut resulted in a lot of images being purged from test registries.  This switches the rhel8 builder to use the canonical location (which was purged) going forward.


grokspawn:
Hmm.... looks like this may not be [necessary anymore](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-operator-framework-olm-main-e2e-gcp-olm-flaky), as the builds of operator-framework-tools seems to be passing now (implying that the image was re-loaded into the registry).

PR #1030, opened 2025-06-30
OCPBUGS-58259: Reduce Frequency of Update Requests for Copied CSVs (#3597)

tmshort:
Use a TransformFunc to generate hashes of the copied CSVs to reduced storage needs, and to minimize API access.

The following is the complete unsquashed commit listing
* (bugfix): reduce frequency of update requests for CSVs

by adding annotations to copied CSVs that are populated with hashes of the non-status fields and the status fields.

This seems to be how this was intended to work, but was not actually working this way because the annotations never actually existed on the copied CSV. This resulted in a hot loop of update requests being made on all copied CSVs.

* update unit tests

* updates to test so far

* Small changes

* Add metadata drift guard to copyToNamespace

Since we switched to a PartialObjectMetadata cache to save memory, we lost visibility into copied CSV spec and status fields, and the reintroduced nonStatusCopyHash/statusCopyHash annotations only partially solved the problem. Manual edits to a copied CSV could still go undetected, causing drift without reconciliation.

This commit adds two new annotations: olm.operatorframework.io/observedGeneration and olm.operatorframework.io/observedResourceVersion. It implements a mechanism to guard agains metadata drift at the top of the existing-copy path in copyToNamespace. If a stored observedGeneration or observedResourceVersion no longer matches the live object, the operator now:

      • Updates the spec and hash annotations
      • Updates the status subresource
      • Records the new generation and resourceVersion in the guard annotations

Because the guard only fires when its annotations are already present, all existing unit tests pass unchanged. We preserve the memory benefits of the metadata‐only informer, avoid extra GETs, and eliminate unnecessary API churn.

Future work may explore a WithTransform informer to regain full object visibility with minimal memory impact.

* Tests for metadata guard

Verifies that exactly three updates (spec, status, guard) are issued when the observedGeneration doesn’t match.

* Persist observed annotations on all status updates

* GCI the file

* Use TransformFunc

Unit tests not updated

* Update operatorgroup tests to compile

* Restore operatorgroup_test from master

Remove metadatalister

* Remove more PartialObjectMetadata

* Remove hashes from operator_test

* Fix error messages for static-analysis

* Update test annotations and test client

* Rename pruning to listerwatcher

* Set resync to 6h

* Add CSV copy revert syncer

* Log tweaks

* Consolidate revert and gc syncers

* Add logging and reduce the amount of metadata in the TransformFunc

* Handle the copy CSV revert via a requeue of the primary CSV

* Revert "Set resync to 6h"

This reverts commit 855f940a2199bd4071c51f14ef44728550bf13cf.

* Add delete handler for copied csv

* Revert whitespace change

* Rename function, fix comment

---------

Upstream-commit: d055f28750cf62f966f566d36990fff5285c7a71 (cherry picked from commit bc111a9c69e1fab45d9696cc10db16cc6fe90ea3)


kuiwang02:
/hold
I will pre-verify it soon. and will unhold it after it is done.

thanks

tmshort:
/retest-required
We seem to be consistently failing:
```
End-to-end: [It] Starting CatalogSource e2e tests when The namespace is labled as Pod Security Admission policy enforce:baseline when A CatalogSource built with opm v1.21.0 (<v1.23.2)is created with spec.GrpcPodConfig.SecurityContextConfig set to legacy The registry pod comes up successfully [CatalogSource] 
```
However, this PR doesn't do anything with CatSrc's...
Going try one more time...

PR #1032, opened 2025-07-07
NO-JIRA: Remove deleted yamls

lance5890:
kind of post-cleanup of https://github.com/openshift/operator-framework-olm/pull/245

address https://github.com/openshift/cluster-version-operator/issues/1207


manifests/0000_50_olm_11-olm-operators.configmap.removed.yaml
```diff
@@ -1,11 +0,0 @@
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: olm-operators
-  namespace: openshift-operator-lifecycle-manager
-  annotations:
-    release.openshift.io/delete: "true"
```
lance5890, line 7: cvo will skip the yaml with the  release.openshift.io/delete: "true" annotation

PR #1056, opened 2025-08-21
(fix) broken generate-manifests script for macOS

anik120:
The script used `realpath --relative-to` which works on Linux (GNU coreutils) but fails on macOS (BSD realpath). This caused `realpath: illegal option -- -` errors when running on macOS.

This PR replaces GNU-specific `realpath --relative-to` with Python's `os.path.relpath()`

Note: This introduces Python3 requirements for both CI environments and dev machines.

Also fixed sed tool selection for script: Use `gsed` (GNU sed) for macOS, `sed` for linux.

### DIFF
--- a/scripts/generate_crds_manifests.sh
+++ b/scripts/generate_crds_manifests.sh
@@ -26,7 +26,7 @@ chartdir="${tmpdir}/chart"
 crddir="${chartdir}/crds"
 crdsrcdir="${tmpdir}/operators"
 
-SED="sed"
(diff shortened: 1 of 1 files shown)


perdasilva (approved):

camilamacedo86:
Closing this one based on: https://github.com/openshift/operator-framework-olm/pull/1056#issuecomment-3448869041
But if you see that is needed pelase feel free to re-open.

/close

PR #1221, opened 2026-02-06
OCPBUGS-76323:OCPQE-31553:remove qe case of testing collect-profiles

kuiwang02:
/cc @Xia-Zhao-rh @bandrade @jianzhangbjz 

the collect-profiles is removed, so the related case need to be deleted.

Assisted-By: Claude Code


jianzhangbjz:
Linked https://github.com/openshift/operator-framework-olm/pull/1215 here.
/approve
/lgtm

PR #1284, opened 2026-04-30
OPRUN-4541,OPRUN-4544: add lifecycle-server for serving FBC catalog lifecycle metadata

perdasilva:
## Summary
- Introduces a new `lifecycle-server` binary that serves lifecycle metadata from FBC (File-Based Catalog) content via a versioned REST API (`GET /api/{version}/lifecycles/{package}`)
- The server loads lifecycle blobs at startup into an in-memory index, serves them over HTTPS with Kubernetes authn/authz, and exposes `/healthz` and `/readyz` endpoints
- Fails fast on FBC load errors and duplicate lifecycle blobs to prevent starting in a degraded or inconsistent state
- Hardens both API and health servers with connection timeouts (ReadHeaderTimeout, ReadTimeout, WriteTimeout, IdleTimeout)
- Includes Dockerfile and Makefile changes for build
- Updates `library-go` to remove cipher suites no longer supported by Go's crypto/tls (DHE-RSA suites, SHA384 CBC suites, `ECDHE-RSA-DES-CBC3-SHA`), which the lifecycle-server's TLS configuration depends on via `crypto.TLSVersion()` and `crypto.CipherSuite()`
- e2es will come in a follow-up PR that adds the lifecycle-controller

## Key Components
- `cmd/lifecycle-server/` — CLI entrypoint with TLS cert hot-reload, graceful shutdown, health/readiness probes, connection timeout hardening
- `pkg/lifecycle-server/` — FBC loading/indexing (`fbc.go`), HTTP API handler and health handler (`server.go`)

## Test plan
- [x] Unit tests for schema version regex matching (`TestSchemaVersionRegex`)
- [x] Unit tests for FBC loading edge cases: missing path, empty dir, mixed schemas, corrupted files, subdirectories, duplicates (`TestLoadLifecycleData`, `TestLoadLifecycleData_Subdirectory`)
- [x] Unit tests for API handler: status codes, content types, method enforcement, nil/empty data, concurrent requests, byte-for-byte blob fidelity (`TestNewHandler*`)
- [x] Unit tests for health/readiness endpoints: healthz always 200, readyz 503 when empty/nil, readyz 200 when loaded (`TestNewHealthHandler`)
- [x] `go build ./cmd/lifecycle-server/...` succeeds
- [x] `go test ./pkg/lifecycle-server/...` passes
- [x] `go mod verify` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a new lifecycle-server with TLS-secured API endpoints and unauthenticated health endpoints to serve lifecycle data by version and package.
  * Introduced in-memory indexing of lifecycle catalog blobs with version/package lookup and readiness semantics.

* **Chores**
  * Build and runtime packaging updated to include the lifecycle-server binary and promote related dependencies.

* **Tests**
  * Extensive unit tests for catalog loading, indexing, HTTP handlers, concurrency, and health/readiness behavior.


pkg/lifecycle-server/server.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"net/http"
+
+	"github.com/go-logr/logr"
+)
+
+// NewHealthHandler creates an HTTP handler for health and readiness probes.
+// The /healthz endpoint always returns 200. The /readyz endpoint returns 200
+// if lifecycle data is loaded, or 503 if the index is empty.
```
perdasilva, line 27: Given that it might be a frequent thing that a catalog does not carry lifecycle metadata, do we want to 503 on /readyz when there's no data?

pkg/lifecycle-server/fbc_test.go
```diff
@@ -0,0 +1,441 @@
...
+		expectedIndex LifecycleIndex
+		expectErr     bool
+	}{
+		{
+			name: "non-existent path returns empty index",
+			setup: func(t *testing.T) string {
+				return filepath.Join(t.TempDir(), "does-not-exist")
+			},
+			expectedIndex: LifecycleIndex{},
+		},
```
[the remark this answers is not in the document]
perdasilva: These findings are incorrect — no fix needed. Here's why:
1. Non-existent path returning empty index is correct behavior. The FBC path might legitimately not exist yet (e.g., the catalog volume   
  hasn't been mounted). Lines 48-51 explicitly handle this by returning an empty index with no error. The caller (start.go) then starts the 
  server, and /readyz reports 503 until data is available. Failing on a missing path would prevent the pod from starting at all.            
  2. Corrupted JSON entries being logged and skipped is correct behavior. The lifecycle-server walks FBC catalogs that contain many schema  
  types (olm.package, olm.channel, etc.) — only io.openshift.operators.lifecycles.* entries are relevant. WalkMetasFS reports per-entry     
  parse errors via the callback's err parameter (line 54-57). Failing the entire load because one unrelated entry in the catalog is
  malformed would be too brittle. The current behavior — log the error, skip the entry, continue loading valid entries — is the right       
  tradeoff. The tests correctly verify this.                           

  The "fail-fast" contract applies to duplicates (data integrity) and top-level walk errors (line 89-91), not to missing paths or individual
   corrupted entries. No changes needed.

perdasilva:
@coderabbitai resume

pkg/lifecycle-server/server.go
fgiudici: Like the idea to get a not 200 HTTP code when no data!
Regarding which code to use, aren't 50x HTTP codes to mean Server errors? 🤔
In this case wouldn't be more appropriate something like 403 (Not Found)?
I have no strong opinions btw.

pkg/lifecycle-server/fbc.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+
+		// Store in index (thread-safe)
+		mu.Lock()
+		if result[schemaVersion] == nil {
+			result[schemaVersion] = make(map[string]json.RawMessage)
+		}
+		if _, exists := result[schemaVersion][meta.Package]; exists {
+			mu.Unlock()
+			return fmt.Errorf("duplicate lifecycle blob for version %q package %q", schemaVersion, meta.Package)
```
fgiudici: nit: we can add also the file path here:
```suggestion
			return fmt.Errorf("duplicate lifecycle blob for version %q package %q in file %q", schemaVersion, meta.Package, path)
```

pkg/lifecycle-server/fbc.go
```diff
@@ -0,0 +1,123 @@
...
+			return nil
+		}
+		schemaVersion := matches[1] // e.g., "v1alpha1"
+
+		if meta.Package == "" {
+			return nil
+		}
+
+		// Store in index (thread-safe)
+		mu.Lock()
```
fgiudici, line 75: nit: what about putting immediately here a 
```
defer mu.Unlock()
```
and drop from line 80 and 84?

fgiudici (commented):
just found a couple of nits
Overall looks great! 🚀

pkg/lifecycle-server/fbc.go
perdasilva: fixed up!

pkg/lifecycle-server/fbc.go
perdasilva: great catch! thank you!

fgiudici (commented):
/lgtm

pkg/lifecycle-server/server.go
perdasilva: I think the problem here is we don't disambiguate between the cases where there was an error parsing the FBC and where there was no data in the first place. I think it could make sense to return a 5xx in the case of errors. If its empty, let's just return 200. I fix this up in a follow up.

fgiudici:
/verified bypass

The lifecycle-server will be spawn by the lifecycle-controller, not yet merged, so this code cannot be really tested.
We are going to test all this new feature only when we will have all the pieces in place.

perdasilva:
overriding failed jobs as this PR doesn't ship anything but an extra binary into the olm container. No manifests, and the binary is never called. Image builds without problems, e2e tests pass without problems.

PR #1288, opened 2026-05-05
OPRUN-4597: Fix lifecycle-server readiness probe to allow empty index

perdasilva:
## Summary
- The readiness probe (`/readyz`) previously returned 503 when the lifecycle index was empty, treating it as unready. An empty index is a valid state (no lifecycle metadata in the catalog), so it now returns 200.
- Returns 503 only when there were actual errors building the index. The server continues to start even on load errors so the readiness probe can report the failure.
- Error logging moved from the readiness handler to FBC processing at startup, so errors are logged once instead of on every probe poll.
- Simplified `NewHealthHandler` to take a `bool` instead of an `error`, keeping the handler decoupled from load internals.

## Test plan
- [x] Unit tests updated and passing (`go test ./pkg/lifecycle-server/...`)
- [ ] Verify readiness returns 200 when catalog has no lifecycle metadata
- [ ] Verify readiness returns 503 with generic message when FBC loading fails
- [ ] Verify error details appear in server logs at startup (not in HTTP response or on each probe poll)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Health & Reliability**
  * Refined health endpoints: liveness stays healthy; readiness now accurately reports degraded state (returns 503) when initialization fails, improving monitoring signals.

* **Bug Fixes**
  * Initialization now logs lifecycle load results and continues startup on load failures while surfacing clearer readiness status and error messages for diagnostics.

### DIFF
--- a/pkg/lifecycle-server/server.go
+++ b/pkg/lifecycle-server/server.go
@@ -24,17 +24,17 @@ import (
 
 // NewHealthHandler creates an HTTP handler for health and readiness probes.
 // The /healthz endpoint always returns 200. The /readyz endpoint returns 200
-// if lifecycle data is loaded, or 503 if the index is empty.
-func NewHealthHandler(data LifecycleIndex) http.Handler {
+// if isReady is true, or 503 otherwise.
+func NewHealthHandler(isReady bool) http.Handler {
 	mux := http.NewServeMux()
 	mux.HandleFunc("GET /healthz", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 		_, _ = w.Write([]byte("ok"))
 	})
 	mux.HandleFunc("GET /readyz", func(w http.ResponseWriter, r *http.Request) {
-		if len(data) == 0 {
+		if !isReady {
 			w.WriteHeader(http.StatusServiceUnavailable)
-			_, _ = w.Write([]byte("no lifecycle data loaded"))
+			_, _ = w.Write([]byte("error processing lifecycle metadata - check logs for details"))
 			return
 		}
 		w.WriteHeader(http.StatusOK)
--- a/cmd/lifecycle-server/start.go
+++ b/cmd/lifecycle-server/start.go
@@ -134,15 +134,16 @@ func run(_ *cobra.Command, _ []string) error {
 
 	// Load lifecycle data from FBC
 	log.Info("loading lifecycle data from FBC", "path", fbcPath)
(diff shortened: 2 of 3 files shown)


pkg/lifecycle-server/server.go
```diff
@@ -17,24 +17,27 @@ limitations under the License.
...
 		w.WriteHeader(http.StatusOK)
 		_, _ = w.Write([]byte("ok"))
 	})
 	mux.HandleFunc("GET /readyz", func(w http.ResponseWriter, r *http.Request) {
-		if len(data) == 0 {
+		if loadErr != nil {
+			log.Error(loadErr, "readiness check failed")
 			w.WriteHeader(http.StatusServiceUnavailable)
-			_, _ = w.Write([]byte("no lifecycle data loaded"))
+			_, _ = w.Write([]byte(fmt.Sprintf("error parsing catalog lifecycle metadata, check logs for details")))
```
fgiudici: We can drop Sprintf here
```suggestion
			_, _ = w.Write([]byte("error parsing catalog lifecycle metadata, check logs for details"))
```

pkg/lifecycle-server/server.go
```diff
@@ -17,24 +17,27 @@ limitations under the License.
...
+func NewHealthHandler(loadErr error, log logr.Logger) http.Handler {
 	mux := http.NewServeMux()
 	mux.HandleFunc("GET /healthz", func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusOK)
 		_, _ = w.Write([]byte("ok"))
 	})
 	mux.HandleFunc("GET /readyz", func(w http.ResponseWriter, r *http.Request) {
-		if len(data) == 0 {
+		if loadErr != nil {
+			log.Error(loadErr, "readiness check failed")
```
fgiudici: nit: logging error here may cause a lot of duplicate warnings in case of errors as the readyz endpoint is polled

fgiudici (commented):
just found a couple of minor things
Looks good!

/lgtm

pkg/lifecycle-server/server.go
perdasilva: great call out!

fgiudici (approved):
/lgtm

fgiudici:
/verified bypass

The lifecycle-server will be spawn by the lifecycle-controller, not yet merged, so this code cannot be really tested.
We are going to test all this new feature only when we will have all the pieces in place.

Issue #1331, opened 2026-06-25
CatalogSource pods should set openshift.io/required-scc annotation to prevent SCC preemption on OpenShift HCP


TGPSKI:
## Summary

On OpenShift HCP (Hosted Control Plane) clusters, CatalogSource pods can be assigned a customer-installed SCC instead of `restricted-v2`, causing the pod to run with an unexpected UID and crash with permission errors on cache directories.

This is actively causing production `CrashLoopBackOff` on ROSA HCP clusters.

## Problem

OpenShift's SCC admission selects the "most restrictive" SCC from the candidate set of the **creating user**. On HCP, the entity that creates catalog pods on the data plane appears to have broader SCC access than the local `catalog-operator` ServiceAccount on classic OpenShift — evidenced by the `security.openshift.io/validated-scc-subject-type: user` annotation on affected pods. This means custom SCCs installed on the cluster can enter the candidate set and win admission.

A custom SCC with `runAsUser.type: MustRunAs` (single fixed UID) scores as more restrictive than `restricted-v2`'s `MustRunAsRange`. When such an SCC wins, it mutates the pod to a UID that differs from the image's expected UID (1001 per `USER` directive in the image Dockerfile), resulting in permission errors on cache directories:

```
open /tmp/cache/pogreb.v1/db/lock: permission denied
```

This does not reproduce on classic OpenShift — the same custom SCC installed on classic causes `restricted-v2` to still win, consistent with the local `catalog-operator` SA not having access to customer-installed SCCs.

Note: `--set-workload-user-id=false` is already set in the OpenShift OLM deployment, so OLM correctly does not inject a `runAsUser`. The UID mutation comes entirely from SCC admission selecting the custom SCC.

## Expected Behavior

CatalogSource pods should consistently run under `restricted-v2` on OpenShift, regardless of what custom SCCs are installed on the cluster.

## Proposed Fix

OpenShift enhancement [openshift/enhancements#1391](https://github.com/openshift/enhancements/pull/1391) (AUTH-372, merged July 2023) introduced the `openshift.io/required-scc` pod annotation, which pins SCC selection to a specific SCC and fails admission hard rather than silently selecting a scored winner.

The fix should be applied in `staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/reconciler.go`, in the `Pod()` function, within the existing `Restricted` security context branch — immediately after the `addSecurityContext` call:

```go
[...]

PR #1332, opened 2026-06-25
fix: pin CatalogSource pods to restricted-v2 SCC on OpenShift HCP

TGPSKI:
## Summary

On ROSA HCP clusters, CatalogSource pods can be assigned a customer-installed SCC instead of `restricted-v2`, causing the pod to run with an unexpected UID and crash with `permission denied` errors on pre-baked cache directories.

This PR adds the `openshift.io/required-scc: restricted-v2` annotation to CatalogSource pods when `securityContextConfig` is `Restricted`, pinning SCC selection and preventing custom SCCs from preempting admission.

## Problem

OpenShift's SCC admission selects the "most restrictive" SCC from the candidate set of the **creating user**. On HCP, the entity that creates catalog pods on the data plane has broader SCC access than the local `catalog-operator` ServiceAccount on classic OpenShift — evidenced by the `security.openshift.io/validated-scc-subject-type: user` annotation on affected pods.

A custom SCC with `runAsUser.type: MustRunAs` (single fixed UID) scores as more restrictive than `restricted-v2`'s `MustRunAsRange`. When such an SCC wins, it mutates the pod to a UID that differs from the image's expected UID (1001 per `USER` directive), resulting in:

```
open /tmp/cache/pogreb.v1/db/lock: permission denied
```

This does not reproduce on classic OpenShift — the same custom SCC installed on classic causes `restricted-v2` to still win, consistent with the local `catalog-operator` SA not having access to customer-installed SCCs.

## Fix

Uses the `openshift.io/required-scc` pod annotation (GA since OCP 4.14, [openshift/enhancements#1391](https://github.com/openshift/enhancements/pull/1391)) to pin SCC selection to `restricted-v2` within the existing `Restricted` security context branch.

Key properties:
- **No-op on non-OpenShift**: The annotation is silently ignored where no SCC admission controller exists
- **Respects user overrides**: `!alreadySet` check preserves any annotation passed through CatalogSource annotations
- **Minimal blast radius**: Only fires in the `Restricted` branch (PSA-labeled namespaces or explicit `grpcPodConfig.securityContextConfig: restricted`)

## Test Plan

- [x] Existing `TestPodExtractContent` tests updated to expect the new annotation in Restricted mode
- [x] New `TestPodRequiredSCCAnnotation` test validates:
  - Restricted config adds the annotation
  - Legacy config does not add the annotation
  - User-provided annotation value is preserved (not overwritten)
- [x] Full `pkg/controller/registry/reconciler/` test suite passes

## References

- Issue: https://github.com/openshift/operator-framework-olm/issues/1331
- Incident: ROSA HCP CrashLoopBackOff
- Enhancement: https://github.com/openshift/enhancements/pull/1391

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/reconciler.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/registry/reconciler/reconciler.go
@@ -356,6 +356,12 @@ func Pod(source *operatorsv1alpha1.CatalogSource, name, opmImg, utilImage, img s
 	if securityContextConfig == operatorsv1alpha1.Restricted {
 		// Apply 'restricted' security settings
 		addSecurityContext(pod, runAsUser)
+
+		// Pin SCC selection to prevent custom SCCs from preempting catalog pods on
+		// HCP, where the creating user has broader SCC access than the local SA.
+		if _, alreadySet := podAnnotations["openshift.io/required-scc"]; !alreadySet {
+			podAnnotations["openshift.io/required-scc"] = "restricted-v2"
+		}
 	}
 
 	// Set priorityclass if its annotation exists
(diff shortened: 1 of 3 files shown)


perdasilva:
@TGPSKI from the PR description:

```
On HCP, the entity that creates catalog pods on the data plane has broader SCC access than the local catalog-operator ServiceAccount on classic OpenShift
```

What is this entity? Could these changes be applied there?

These changes would need to be made in the upstream and then get downsync'ed, but we are a bit reluctant to add OCP specific details to the upstream...

TGPSKI:
> What is this entity? Could these changes be applied there?

Provided response in private channel

tmshort:
/hold
As pointed out, the contents of the `staging` directory need to be made in the upstream repos.

tmshort:
@TGPSKI please move this work upstream.

PR #1335, opened 2026-06-29
TRT-2761: Undo revert and fix applyClusterTLSProfile()

tmshort:
This undoes the revert #1334 (which reverted #1330). The downstream sync is pulled back in, and the fix https://github.com/operator-framework/operator-lifecycle-manager/pull/3860 is downstreamed in. This should be the only necessary fix, and the only commit that needs to be downstreamed.

This has to be a manual downstream due to the revert. Following this, subsequent automatic syncing should succeed.

Alternative to #1333 

## Summary by CodeRabbit

* **New Features**
  * Package server now can automatically adopt the cluster’s TLS security profile when no custom TLS minimum is set.
  * Added permission to read the cluster APIServer configuration needed for this startup behavior.

* **Bug Fixes**
  * Improved handling so existing TLS settings provided by users continue to take precedence.
  * Added broader compatibility and stability updates through dependency and workflow upgrades.


stbenjam:
/payload-job periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn

PR #1341, opened 2026-07-31
OCPBUGS-86895: Ensure packageserver pod seccompProfile is always set

tmshort:
## Summary

- The `openshift-operator-lifecycle-manager` namespace enforces `restricted:latest` PodSecurity, requiring `seccompProfile: RuntimeDefault` on all pods.
- The packageserver CSV template includes this field, but the cluster-stored CSV can diverge (created before the field was added, or via an upgrade race where OLM processes a cached pre-update CSV).
- OLM faithfully reproduces whatever is in the cluster CSV, so the generated Deployment pod template is missing `seccompProfile`. On most clusters the `restricted-v2` SCC mutating admission adds it before PSA validates — masking the bug. On OSD/ROSA Classic, PSA runs before SCC mutation, so pods are rejected and OLM enters a perpetual reinstall loop (observed at generation=716, revision=566).
- Fix: explicitly enforce `seccompProfile: RuntimeDefault` in `ensureCSVHighAvailability`, which runs on every PSM reconcile, making the field mandatory regardless of the stored CSV state.

## Test plan

- [x] Three new unit test cases added to `TestEnsureCSV`: missing `seccompProfile` (HA topology), missing `seccompProfile` (single-replica topology), nil `SecurityContext`
- [x] `go test ./pkg/package-server-manager/...` passes
- [ ] Backport to all active release branches (4.18, 4.19, 4.20, ...)

Fixes: https://issues.redhat.com/browse/OCPBUGS-86895

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment security by ensuring pods use the `RuntimeDefault` seccomp profile.
  * Automatically creates or restores missing pod security configuration.
  * Applies these corrections consistently to highly available and single-replica deployments.
  * Deployments are now marked for update when security settings require correction.

### DIFF
--- a/pkg/package-server-manager/config.go
+++ b/pkg/package-server-manager/config.go
@@ -149,6 +149,21 @@ func ensureCSVHighAvailability(image string, csv *olmv1alpha1.ClusterServiceVers
 		modified = true
 	}
 
+	// Ensure pod-level seccompProfile is always set. The openshift-operator-lifecycle-manager
+	// namespace enforces restricted:latest PodSecurity, which requires seccompProfile. The CSV
+	// template includes this field, but it may be absent if the cluster CSV was created by an
+	// older version or if SCC/PSA admission ordering masks the missing field on most clusters.
+	if deployment.Template.Spec.SecurityContext == nil {
+		deployment.Template.Spec.SecurityContext = &corev1.PodSecurityContext{}
+	}
+	if deployment.Template.Spec.SecurityContext.SeccompProfile == nil ||
+		deployment.Template.Spec.SecurityContext.SeccompProfile.Type != corev1.SeccompProfileTypeRuntimeDefault {
+		deployment.Template.Spec.SecurityContext.SeccompProfile = &corev1.SeccompProfile{
+			Type: corev1.SeccompProfileTypeRuntimeDefault,
+		}
+		modified = true
+	}
+
 	if modified {
 		csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec = *deployment
 	}
--- a/pkg/package-server-manager/controller_test.go
+++ b/pkg/package-server-manager/controller_test.go
@@ -94,6 +94,21 @@ func withAffinity(affinity *corev1.Affinity) func(*olmv1alpha1.ClusterServiceVer
 		csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec.Template.Spec.Affinity = affinity
 	}
 }
+
+func withoutSeccompProfile() testCSVOption {
+	return func(csv *olmv1alpha1.ClusterServiceVersion) {
+		podSpec := &csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec.Template.Spec
+		if podSpec.SecurityContext != nil {
+			podSpec.SecurityContext.SeccompProfile = nil
+		}
+	}
+}
+
+func withNilPodSecurityContext() testCSVOption {
+	return func(csv *olmv1alpha1.ClusterServiceVersion) {
+		csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec.Template.Spec.SecurityContext = nil
+	}
+}
 func withRollingUpdateStrategy(strategy *appsv1.RollingUpdateDeployment) func(*olmv1alpha1.ClusterServiceVersion) {
 	return func(csv *olmv1alpha1.ClusterServiceVersion) {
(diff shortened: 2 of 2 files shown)


pkg/package-server-manager/controller_test.go
```diff
@@ -255,6 +270,27 @@ func TestEnsureCSV(t *testing.T) {
...
+			inputCSV:        newTestCSV(withReplicas(singleReplicas), withRollingUpdateStrategy(emptyRollout), withAffinity(&corev1.Affinity{}), withoutSeccompProfile()),
+			expectedCSV:     newTestCSV(withReplicas(singleReplicas), withRollingUpdateStrategy(emptyRollout), withAffinity(&corev1.Affinity{})),
+		},
+		{
+			name:            "Modified/HighlyAvailable/NilPodSecurityContext",
+			want:            wanted{true, nil},
+			highlyAvailable: true,
+			inputCSV:        newTestCSV(withReplicas(defaultReplicas), withRollingUpdateStrategy(defaultRollout), withAffinity(defaultAffinity), withNilPodSecurityContext()),
+			expectedCSV:     newTestCSV(withReplicas(defaultReplicas), withRollingUpdateStrategy(defaultRollout), withAffinity(defaultAffinity)),
+		},
```
hongkailiu, line 293: Do we want to have a case of "Modified/SingleReplica/NilPodSecurityContext"?

My guess is the intension is that HA is more important here and for SingleReplica we do not need as much coverage. A case of "Modified/SingleReplica/MissingSeccompProfile" seems enough.

pkg/package-server-manager/config.go
```diff
@@ -149,6 +149,21 @@ func ensureCSVHighAvailability(image string, csv *olmv1alpha1.ClusterServiceVers
...
 
+	// Ensure pod-level seccompProfile is always set. The openshift-operator-lifecycle-manager
+	// namespace enforces restricted:latest PodSecurity, which requires seccompProfile. The CSV
+	// template includes this field, but it may be absent if the cluster CSV was created by an
+	// older version or if SCC/PSA admission ordering masks the missing field on most clusters.
+	if deployment.Template.Spec.SecurityContext == nil {
+		deployment.Template.Spec.SecurityContext = &corev1.PodSecurityContext{}
+	}
+	if deployment.Template.Spec.SecurityContext.SeccompProfile == nil ||
+		deployment.Template.Spec.SecurityContext.SeccompProfile.Type != corev1.SeccompProfileTypeRuntimeDefault {
```
hongkailiu, line 160: (probably because of my limited knowledge on `SecurityContext`) I do not see why we enforce `corev1.SeccompProfileTypeRuntimeDefault`.
But it looks like a natural choice out of the 3 candidates.

https://github.com/openshift/operator-framework-olm/blob/5bebe93fb8459eb3b5ad94a6edf3895af0005ab6/vendor/k8s.io/api/core/v1/types.go#L5006-L5014

hongkailiu (commented):
/lgtm

The fix looks good to me.
I left a couple of questions for my own learning.

pkg/package-server-manager/config.go
tmshort: I think that's basically the reason; we don't want to rely on `Localhost` or `Unconfined`

pkg/package-server-manager/controller_test.go
tmshort: It might be something to consider in the future.

hongkailiu: :resolved.

pkg/package-server-manager/config.go
hongkailiu: :resolved. Thanks for the answers.

vwalek: The "RuntimeDefault" is usually added when the Object is processed by the controller. By default, any pod running under `restricted-v2` should have the seccompProfile set to `RuntimeDefault`. However, it is better to explicitly define it in the deployment as we always want to have the default value.
Just adding my thoughts about the issue.

tmshort: It is the intent to explicitly define the setting.

rashmigottipati (commented):
/verified by ci

Leo6Leo:
Manual Verification Results

Cluster: OCP 5.0.0-0.nightly-2026-08-04-023110

On-Cluster Verification

1. Removed seccompProfile from the packageserver CSV — the controller reconciled immediately and restored seccompProfile: RuntimeDefault. 2. Removed the entire securityContext from the CSV — the controller recreated it with runAsNonRoot: true and seccompProfile: RuntimeDefault. 3. Pod health — packageserver pods remained Running with 0 restarts throughout both tests.

/verified by @Leo6Leo

PR #1347, opened 2026-08-12
OCPBUGS-35210: prevent KCM from deleting SA-token secrets created before their SA

rsacherer:
SA-token Secrets included in an operator bundle are placed earlier in the InstallPlan step list than the synthesized ServiceAccount step. OLM creates the Secret before the SA exists; the Kubernetes token controller (KCM) immediately deletes any token secret whose referenced ServiceAccount is absent, and OLM then marks the step as Created permanently — preventing any future retry and leaving the operator without its token secret.

The issue is not always reproduceable (see test document) because a rescheduled loop run is able to get the stale RV, which still does not see the Secret as created, but the prior loop created the SA so this time the Secret will not be deleted.

Fix: add a StepperFunc for BundleSecretKind (mirroring the existing CRD StepperFunc pattern). The new NewBundleSecretStep checks whether the SA referenced by the secret already exists before attempting creation:

- SA absent  → return WaitingForAPI; NeedsRequeue() returns true, keeping phase=Installing and triggering a 5-second requeue.
- SA present → create the secret with correct owner refs (live API UID lookup, matching getUpdatedOwnerReferences behaviour) and return Created/Present.

Because the StepperFunc handles WaitingForAPI internally it never reaches the main ExecutePlan switch case that would otherwise skip the step, so no changes to the switch statement or to NeedsRequeue() are required.

Also adds structured debug logging to syncInstallPlans (plan resourceVersion, per-step BS/SA status at reconcile start, UpdateStatus call/result) to make the race observable in OLM pod logs during investigation.

Most likely the additional logging will be removed via further commits.

Tested: 30-iteration statistical reproducer against OCP 4.17 — 0/30 bug fires with the fix (without the fix we are looking at roughly 60% failure/40% Success).

This PR is still a draft and work in progress.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved InstallPlan processing for BundleSecret resources, including creation, updates, labeling, and ownership tracking.
  * Installations now better handle service-account-token Secrets and wait for required service accounts when needed.
  * Improved step status handling and reporting during installation, helping prevent incomplete or inconsistent resource setup.

* **Diagnostics**
  * Added more precise timestamps and detailed installation progress logging to support troubleshooting.

### DIFF
--- a/staging/operator-lifecycle-manager/cmd/catalog/start.go
+++ b/staging/operator-lifecycle-manager/cmd/catalog/start.go
@@ -49,6 +49,16 @@ func newRootCmd() *cobra.Command {
 			if o.debug {
 				logger.SetLevel(logrus.DebugLevel)
 			}
+			// OCPBUGS-35210: use millisecond timestamps so OLM and audit log
+			// entries can be correlated at sub-second precision.
+			// Set on both the local logger AND the global package logger so that
+			// code using logrus.WithFields() directly also emits milliseconds.
+			msFormatter := &logrus.TextFormatter{
+				TimestampFormat: "2006-01-02T15:04:05.000000Z07:00",
+				FullTimestamp:   true,
+			}
+			logger.SetFormatter(msFormatter)
+			logrus.SetFormatter(msFormatter)
 			logger.Infof("log level %s", logger.Level)
 
 			ctx, cancel := context.WithCancel(signals.Context())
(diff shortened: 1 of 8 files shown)


tmshort:
/hold
This modifies the `staging` directory, so the changes needs to be moved upstream.

PR #1348, opened 2026-08-12
fix(olm): fix ConversionWebhook spec.conversion lifecycle during CSV upgrades

ugiordan:
### 1. spec.conversion written before pods are ready (upgrade race)

During an OLM-managed upgrade, `spec.conversion` is written to CRDs during `Install()` — specifically inside `installDeployments()` → `createOrUpdateCertResourcesForDeployment()` — before any pod from the new deployment is scheduled or ready.

Once `spec.conversion` is set, the apiserver routes conversion calls to the new webhook service endpoint. But since no new pod is serving `/convert` yet, those calls return HTTP 404. This breaks CRD version conversion mid-upgrade and causes upgrade failures.

Empirical evidence from testing: `manager=catalog` writes `spec.conversion` at T+0 (no caBundle), `manager=olm` overwrites at T+8s (adds caBundle, points to new service) — both happen before any new pod is ready.

This is the root cause of the RHOAI 3.5 GA blocker RHOAIENG-76183.

### 2. spec.conversion not cleared when replacement CSV drops the ConversionWebhook

`handleClusterServiceVersionDeletion` returned unconditionally when any replacement CSV was found, assuming the replacement would manage `spec.conversion` going forward. If the replacement dropped the ConversionWebhook entirely, `spec.conversion` was left pointing at the now-deleted service. All CR conversion requests then fail with connection refused.

### Fix 1: defer spec.conversion write until deployment is ready

Skip `ConversionWebhook` descriptors in `createOrUpdateCertResourcesForDeployment()` so that `spec.conversion` is never written during `Install()`.

Add `EnsureConversionWebhooks()` on `*StrategyDeploymentInstaller` and call it from `areWebhooksAvailable()`, which is only invoked from `updateInstallStatus()` after `CheckInstalled()` confirms the deployment's pods are ready. Gate `areWebhooksAvailable()` behind `strategyInstalled && strategyErr == nil` so `EnsureConversionWebhooks()` is never called before readiness is confirmed.

#### Why not extend the StrategyInstaller interface?

`EnsureConversionWebhooks()` is intentionally a concrete method on `*StrategyDeploymentInstaller` rather than an interface method. Adding it to `StrategyInstaller` would require updating `NullStrategyInstaller` and all generated counterfeiter fakes. The `areWebhooksAvailable()` call site type-asserts to `*StrategyDeploymentInstaller` before calling the method — a safe assert since `NullStrategyInstaller` never has `ConversionWebhook` entries.

### Fix 2: clear spec.conversion for CRDs dropped by the replacement CSV

Instead of returning unconditionally when a replacement CSV is found, build the set of CRDs still covered by a ConversionWebhook in the replacement. Only reset `spec.conversion` to `NoneConverter` for CRDs the new CSV dropped. CRDs the replacement still covers are left intact so in-flight conversion calls keep working during a normal upgrade.

Also adds a nil guard on `crd.Spec.Conversion` before writing to it, fixing a latent panic in the no-replacement path.

## Files changed

- `staging/operator-lifecycle-manager/pkg/controller/install/deployment.go`: skip ConversionWebhook in `createOrUpdateCertResourcesForDeployment()`, add `EnsureConversionWebhooks()`
- `staging/operator-lifecycle-manager/pkg/controller/operators/olm/apiservices.go`: add `installer` param to `areWebhooksAvailable()`, call `EnsureConversionWebhooks()` before checking CRD state
- `staging/operator-lifecycle-manager/pkg/controller/operators/olm/operator.go`: gate `areWebhooksAvailable()` behind `strategyInstalled && strategyErr == nil`; fix `handleClusterServiceVersionDeletion` to only clear `spec.conversion` for CRDs the replacement CSV dropped

## Testing

Both packages compile cleanly:
```
go build ./staging/operator-lifecycle-manager/pkg/controller/install/...
go build ./staging/operator-lifecycle-manager/pkg/controller/operators/olm/...
```

Companion operator-side fix: [opendatahub-io/opendatahub-operator#3958](https://github.com/opendatahub-io/opendatahub-operator/pull/3958) removes `ConversionWebhook` from the bundle CRD manifests and adds a runtime Runnable to patch `spec.conversion` after the webhook server is confirmed ready.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversion webhook installation by applying configuration after deployments are ready.
  * Added readiness checks and retry handling when conversion webhooks are unavailable.
  * Prevented unnecessary removal of conversion webhooks during CSV upgrades and deletions.
  * Improved cleanup safety when replacement resources exist, listings fail, or conversion settings are missing.
  * Install status now validates webhook availability only after a successful installation.

### DIFF
--- a/staging/operator-lifecycle-manager/pkg/controller/install/deployment.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/install/deployment.go
@@ -123,12 +123,36 @@ func (i *StrategyDeploymentInstaller) createOrUpdateCertResourcesForDeployment()
 				return err
 			}
 		case *webhookDescriptionWithCAPEM:
+			// ConversionWebhook CRD patching is deferred to the post-Install readiness
+			// check (areWebhooksAvailable) so that spec.conversion is only written once
+			// the new deployment's pods are actually serving /convert. Writing it here
+			// during Install() — before any pod is ready — causes a window where the
+			// apiserver routes conversion calls to pods that return HTTP 404.
+			if d.webhookDescription.Type == v1alpha1.ConversionWebhook {
+				continue
+			}
 			err := i.createOrUpdateWebhook(d.caPEM, d.webhookDescription)
 			if err != nil {
 				return err
 			}
 		default:
 			return fmt.Errorf("unsupported CA Resource")
+		}
+	}
+	return nil
+}
+
+// EnsureConversionWebhooks writes spec.conversion on CRDs for any ConversionWebhook
+// entries in the cert resources. Called after the deployment is confirmed ready so that
+// the conversion endpoint is only activated when the new pods are serving /convert.
+func (i *StrategyDeploymentInstaller) EnsureConversionWebhooks() error {
+	for _, desc := range i.getCertResources() {
+		d, ok := desc.(*webhookDescriptionWithCAPEM)
+		if !ok || d.webhookDescription.Type != v1alpha1.ConversionWebhook {
+			continue
+		}
+		if err := i.createOrUpdateConversionWebhook(d.caPEM, d.webhookDescription); err != nil {
+			return err
 		}
 	}
 	return nil
--- a/staging/operator-lifecycle-manager/pkg/controller/operators/olm/apiservices.go
+++ b/staging/operator-lifecycle-manager/pkg/controller/operators/olm/apiservices.go
@@ -562,7 +562,11 @@ func (a *Operator) cleanUpRemovedWebhooks(csv *v1alpha1.ClusterServiceVersion) e
 	return nil
 }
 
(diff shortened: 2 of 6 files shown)


tmshort:
/hold
This includes changes in the `staging` directory; these files need to be modified upstream first (operator-framework/operator-lifecycle-manager), then pulled downstream.

ugiordan:
> /hold This includes changes in the `staging` directory; these files need to be modified upstream first (operator-framework/operator-lifecycle-manager), then pulled downstream.

@tmshort https://github.com/operator-framework/operator-lifecycle-manager/pull/3892

grokspawn:
As Todd mentions, content from the staging directories downstream comes from upstream changes which the project synchs downstream.  I'm closing this to reduce the chance that there is confusion later on provenance.
/close

PR #1359, opened 2026-08-25
OCPBUGS-83412: set catalogsource spec.grpcpodconfig.scc: restricted for all non-legacy cases

grokspawn:
Set explicit SCC for all catsrc specs used for non-legacy cases (defined as where the opm version inside does not support PSA labeling concerns for OCP... pre 1.21), since this can cause a PodSecurityViolation event to get logged.
Since these specs are used in namespaces other than openshift-* naming but do not actually test legacy pre-PSA-support behavior, best practice is to define the SCC so that OLM will set appropriate labels to satisfy OCP PSA audits.

## Summary by CodeRabbit

- **Security**
  - CatalogSource gRPC pods now use a restricted security context by default across supported deployment templates.
  - Improves workload isolation and aligns CatalogSource deployments with stricter security requirements.

- **Tests**
  - Updated CatalogSource test scenarios to validate deployments using the restricted security configuration.

### DIFF
--- a/tests-extension/pkg/bindata/qe/bindata.go
+++ b/tests-extension/pkg/bindata/qe/bindata.go
@@ -359,6 +359,7 @@ objects:
   spec:
     image: "${ADDRESS}"
     grpcPodConfig:
+      securityContextConfig: restricted
       extractContent:
         catalogDir: /configs
     secrets:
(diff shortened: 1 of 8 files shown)


grokspawn:
/retest
(not sure if my previous command worked or was ignored, but there's lots of pressure on app.ci right now, so I'll belt-and-suspenders this command)

tmshort:
/lgtm
Some whitespace changes, but they're fixing things.

